PR #40, opened 2023-04-11
chore: added storybook as a part of scaffolding

Usama-Mavric:


packages/plugin/src/generators/app/templates/src/stories/Button.stories.ts
actong, line 1: Why are there stories for Button, Header, and Page?  I'd think the only story would be for the sample widget.

rverma3: @Usama-Mavric The plan is to have a "sample" widget in the src folder. It will have the same folder structure like that of our widgets in the "widgets" codebase, along with the properties.json file. 

- Maybe to start with you can create a very simple echo widget -- All this component can do is take a message:string as prop and displays that message.
- In the properties.json file, you will have just one property of type "text" and name "message" (this generated folder is supposed to work exactly similar to how our "widgets" repo works.)
- In the stories folder, we can have just one Sample.stories.tsx which will call our sample widget

Usama-Mavric: Done.

rverma3 (approved):

PR #41, opened 2023-04-12
CDR-631

bradwray:
Added manifest generating script, removed inline source mapping, added global access to .env variables and an env.sample file.


bradwray:
The failing of that lint check is because of the use of assert inside of the import on line 13 of generate-manifest.js. That method is how we do it in the widgets library. The terminal says this about it...
"ExperimentalWarning: Importing JSON modules is an experimental feature. This feature could change at any time"

Trying to disable that eslint rule has not been successful. Suggestions? Perhaps we don't assert for type json there?

package-lock.json
actong, line 1: What changed here?  Doesn't seem like there were any changes that should have affected the lock file.

packages/plugin/src/generators/app/templates/.env.sample
```diff
@@ -0,0 +1,3 @@
+//You'll need to create a .env file with the following variables
+
+REACT_APP_API_ROOT=<<api url string>>
```
actong, line 3: Where would this be used?

packages/plugin/src/generators/app/templates/scripts/generate-manifest.js
actong, line 1: This should really go in plugin-scripts.

packages/plugin/src/generators/app/templates/webpack.config.js
```diff
@@ -70,5 +71,11 @@ module.exports = {
         new CopyWebpackPlugin({
             patterns: ['manifest.json'],
         }),
+        new ProvidePlugin({
+            process: 'process/browser',
```
actong, line 75: Is this to provide `process` in the browser?  Why is that needed?

packages/plugin/src/generators/app/templates/manifest.json
```diff
@@ -7,7 +7,7 @@
             "name": "Sample widget",
             "description": "Description of the Sample Widget",
             "version": "1.0.0",
-            "src": "src/widgets/Widget",
+            "src": "src/widgets/SampleWidget",
```
actong, line 10: This change itself is fine, though with generate-manifest, we wouldn't include this file at all anymore right?  It probably should be added to .gitignore, or better yet, generate-manifest should just generate the manifest directly in the `dist` directory as a build artifact instead of in the source tree and then relying on CopyWebpackPlugin to copy it to the build output.

packages/plugin/src/generators/app/templates/webpack.config.js
rverma3: Yes, this resolves the issue of not being able to access environment variables using process.env.<<var>>

actong: Is there a scenario where we need process.env?  My feeling is there shouldn't be a need for using process.env anywhere in widgets.

rverma3: Currently there isn't. I just thought it'll be useful to include support for it incase they need something for their local code development outside of what the context lib is providing. And they needn't have to fiddle with the configuration.

packages/plugin/src/generators/app/templates/.env.sample
rverma3: Yes, this can be removed.

packages/plugin/src/generators/app/templates/webpack.config.js
actong: Well, the reason process/browser even exists (btw I assume for this to work we'd need to include 'process' as a dependency in package.json) is because there is no `process` or `process.env` in the browser.  Even if you include this, there wouldn't be anything in `process.env` when running in the browser, so I'm not sure what this would be used for.  If this is stemming from things like `STORYBOOK_*` environment variables, those are used at build time so it wouldn't need this plugin anyway.

packages/plugin/src/generators/app/templates/scripts/generate-manifest.js
bradwray: When generating the manifest it currently adds "name: '<%= projectName %>'," to the manifest. Is this not needed? Is it ok to be generic? If we move it outside of the yeoman template it won't be able to get that value.

packages/plugin/src/generators/app/templates/.env.sample
bradwray: I spoke with @rverma3. I'll remove this

package-lock.json
bradwray: I used npm i instead of yarn install. That would do it, right?

packages/plugin/src/generators/app/templates/scripts/generate-manifest.js
actong: You can read the project name from the project's package.json.

package-lock.json
actong: Yeah, this repo uses yarn so this file should be removed.

bradwray: Should probably be mentioned in the readme

packages/plugin/src/generators/app/templates/webpack.config.js
rverma3: That's what I'm trying to think through..., they _could_ add things in process.env (yes, I agree `process` dependency will need to be added) for their code -- we can't predict what will be there in the widgets component code. (I guess I;m not sure that we can safely assume that there will _never_ be anything in the `process.env` when running in the browser)

packages/plugin/src/generators/app/templates/scripts/generate-manifest.js
```diff
@@ -0,0 +1,47 @@
...
+    const folderName = 'src/widgets';
+    const getWidgetProperties = async (widgetsList) => {
+        const temp = [];
+        console.log('');
+        for await (const widget of widgetsList) {
+            try {
+                const widget_folder_path = folderName + '/' + widget;
+                // eslint-disable-next-line node/no-unsupported-features/es-syntax
+                let WidgetProperties = await import('../' + widget_folder_path + '/WidgetProperties.json', {
+                    assert: { type: 'json' },
```
rverma3, line 14: @bradwray  is this what you meant to remove in your suggestion to fix the warning? I don't remember for sure if removing this assertion will still work (i.e. will the json still be imported or not). But yea, try and see if it does.

packages/plugin/src/generators/app/templates/webpack.config.js
actong: You can be sure there will never be anything in `process.env` because `process` doesn't even exist in the browser.  This code is polyfilling a global `process`, but there will be nothing in `process.env`, it'll be empty.  There's really no reason for widgets to be using `process`, and if for some reason someone does want a global `process` variable in the browser, it'd really be up to them to set that up.

> Also -- im curious -- when you mentioned STORYBOOK-* vars are used at build time -- won't those be a part of process.env?

If you mean at runtime in the browser, no.  There is no `process.env` in the browser.  If you try to use it, you'll get a 'process is undefined' error.  The webpack build replaces the `process.env.*` references to whatever value is in the environment at the time of the build.  The output of webpack doesn't have `process.env`.

packages/plugin/src/generators/app/templates/scripts/generate-manifest.js
bradwray: Yeah that comma after the file path shows a parsing error. And the terminal complains that running assertions on an import is an experimental feature.

rverma3: @bradwray i meant we should check if we remove the assert, will the import still work as expected for jsons? If it does, im fine with removing the assertion.

packages/plugin-scripts/bin/generate-manifest.js
actong, line 1: Can you add the file header with the copyright?  E.g. see package-plugin.js.

Note, I'm not sure if vscode is suggesting an extension to install, but you can install the psioniq file header extension, which should take care of adding these file headers automatically, though since this file already exists, you'll have to manually do it (or use the vscode command palette and run "Header Insert").  Note though, that since this is a script file, the `#!/usr/bin/env` should be at the top, so the file header would actually go below that.

packages/plugin/src/generators/app/templates/webpack.config.js
```diff
@@ -69,10 +68,7 @@ module.exports = {
             ],
         }),
         new CopyWebpackPlugin({
```
actong, line 70: I don't think this is necessary anymore now that manifest.json is in ./dist.

packages/plugin/src/generators/app/templates/webpack.config.js
```diff
@@ -68,7 +68,10 @@ module.exports = {
...
-            patterns: ['manifest.json'],
+            patterns: ['./dist/manifest.json'],
+        }),
+        new Dotenv({
```
actong, line 73: Can this be removed as well now or it's still needed for something?

packages/plugin-scripts/bin/generate-manifest.js
```diff
@@ -1,21 +1,25 @@
...
-                // eslint-disable-next-line node/no-unsupported-features/es-syntax
-                let WidgetProperties = await import('../' + widget_folder_path + '/WidgetProperties.json', {
-                    assert: { type: 'json' },
-                });
-                WidgetProperties.default.src = widget_folder_path;
-                temp.push(WidgetProperties.default);
-                console.info(WidgetProperties.default.name + ' widget -- READ');
+                const widget_folder_path = path.join(folderName, widget);
+                const widgetProps = require(`${widget_folder_path}/WidgetProperties.json`);
+                let WidgetProperties = JSON.parse(JSON.stringify(widgetProps));
```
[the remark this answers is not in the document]
bradwray: The previous code (brought in from the widgets repo) did an assertion to check that it was valid json. This serves the same purpose while also passing that eslint check that we have on this repo. If it is not valid it'll get caught in the catch.

packages/plugin/src/generators/app/templates/webpack.config.js
bradwray: Just to be clear. You're saying that the CopyWebpackPlugin would just be saving maniifest.json into the dist folder where it already lives, right?

actong: Yeah, I assume that's what it would be doing.  In any case, the manifest is already in ./dist, which is all we really need (`pkgplugin` takes everything in ./dist and zips it up), so we shouldn't need to do anything else.

packages/plugin-scripts/bin/generate-manifest.js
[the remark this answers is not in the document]
actong: This won't validate the file contains valid JSON.  JSON is a string representation of some javascript value.  If what was in WidgetProperties.json wasn't valid JSON, it probably wouldn't even get here, and even if it did, the JSON has already been parsed.  At this point `widgetProps` is already referencing a valid javascript value, i.e. it's no longer JSON, and stringifying it to JSON and parsing it back will just yield the same javascript value.  This is essentially a no-op.  I think the only reason it could fail is if `widgetProps` contained circular references, in which case `stringify` might fail.

packages/plugin/src/generators/app/templates/package.json
```diff
@@ -28,6 +28,7 @@
         "@types/react": "^18.0.12",
         "@types/react-dom": "^18.0.5",
         "babel-loader": "^9.1.2",
+        "dotenv-webpack": "^8.0.1",
         "copy-webpack-plugin": "^11.0.0",
```
actong, line 32: Both of the above plugins can be removed now.

packages/plugin-scripts/bin/generate-manifest.js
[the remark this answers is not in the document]
bradwray: Thanks, that answers this question I asked yesterday https://github.com/Evoke-Platform/evoke-sdk/pull/41#issuecomment-1505662676

actong (commented):
BTW, is commitlint running when you do a commit?  I noticed your last commit followed conventional commits (e.g. chore:), but the prior ones didn't.

PR #51, opened 2023-04-17
chore: rename plugin to create-widgets-package

bradwray:
Story also specifies that the template should contain no private dependencies. That has been confirmed as well.


bradwray:
Renaming this is the story's requirement, right?

actong:
I guess I'm questioning the story then.  I'd want to avoid restricting this to just widgets.  Right now we only support widgets but this may expand in the future.  In the services apis, the uploaded zips are referred to as plugins.

actong:
Closing obsolete PR.

PR #61, opened 2023-04-24
Widget tsdoc tags

actong:
This is a proposal for a different method of describing widgets that doesn't involve editing a separate json file.  Instead the widget description is specified by using tsdoc @tags inline with the respective code.   A minimal description would just mark the widget function with the `@widget` tag:

```typescript
/** @widget */
function MyWidget(props: { title: string }) {
  ...
}
```

The manifest generation will detect the tsdoc tag and generate the following description:

```json
{
  "id": "MyWidget",
  "name": "MyWidget",
  ...
  "properties": [
    {
      "name": "title",
      "type": "text",
      ...
    }
  ]
}
```

Additional tags (`@widgetName`, `@propertyName`, `@propertyType`, etc.) can be used to fine tune the widget description.

This is not fully feature complete, but it should work for any widget that only requires simple props.  WidgetProperties.json can still be used instead, and any existing widgets making use of WidgetProperties.json will continue to use the json file.

### DIFF
--- a/packages/plugin-scripts/src/manifest/descriptors.ts
+++ b/packages/plugin-scripts/src/manifest/descriptors.ts
@@ -1,0 +1,19 @@
+// Copyright (c) 2023 System Automation Corporation.
+// This file is licensed under the MIT License.
+
+export type WidgetDescriptor = {
+    id: string;
+    name: string;
+    description?: string;
+    version?: string;
+    src: string;
+    properties: WidgetPropertyDescriptor[];
+};
+
+export type WidgetPropertyDescriptor = {
+    name: string;
+    displayName: string;
+    type: string;
+    optional?: boolean;
+    [key: string]: unknown;
+};
--- a/packages/plugin-scripts/src/manifest/descriptors.ts
+++ b/packages/plugin-scripts/src/manifest/descriptors.ts
@@ -12,8 +12,8 @@ export type WidgetDescriptor = {
 
 export type WidgetPropertyDescriptor = {
     name: string;
-    displayName: string;
+    displayName?: string;
     type: string;
-    optional?: boolean;
+    isOptional?: boolean;
     [key: string]: unknown;
 };
--- a/packages/plugin-scripts/src/manifest/parser.ts
+++ b/packages/plugin-scripts/src/manifest/parser.ts
@@ -1,0 +1,175 @@
+// Copyright (c) 2023 System Automation Corporation.
+// This file is licensed under the MIT License.
+
+import * as tsdoc from '@microsoft/tsdoc';
+import { Block, parse as parseComment } from 'comment-parser';
+import ts, { SourceFile, SyntaxKind, TypeAliasDeclaration } from 'typescript';
+import debug from './debug';
+
+// Adapted from @microsoft/tsdoc api-demo
+// https://github.com/microsoft/tsdoc/tree/main/api-demo
+
+export interface Comment {
+    compilerNode: ts.Node;
+    commentBlock: Block;
+}
+
+export type ParsedSource = {
+    sourceFile: SourceFile;
+    comments: Comment[];
+    types: {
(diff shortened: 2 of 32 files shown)


packages/plugin-scripts/src/manifest/descriptors.ts
```diff
@@ -0,0 +1,19 @@
...
+    version?: string;
+    src: string;
+    properties: WidgetPropertyDescriptor[];
+};
+
+export type WidgetPropertyDescriptor = {
+    name: string;
+    displayName: string;
+    type: string;
+    optional?: boolean;
```
rverma3, line 17: the property-name is **isOptional**

packages/plugin-scripts/src/manifest/descriptors.ts
```diff
@@ -0,0 +1,19 @@
...
+    name: string;
+    description?: string;
+    version?: string;
+    src: string;
+    properties: WidgetPropertyDescriptor[];
+};
+
+export type WidgetPropertyDescriptor = {
+    name: string;
+    displayName: string;
```
rverma3, line 15: This is an optional property.

packages/plugin-scripts/src/manifest/parser.ts
```diff
@@ -0,0 +1,175 @@
...
+};
+
+/**
+ * Returns true if the specified SyntaxKind is part of a declaration form.
+ *
+ * Based on ts.isDeclarationKind() from the compiler.
+ * https://github.com/microsoft/TypeScript/blob/v3.0.3/src/compiler/utilities.ts#L6382
+ */
+function isDeclarationKind(kind: ts.SyntaxKind): boolean {
+    return (
```
rverma3, line 32: @actong curious. what is this for?

actong: This comes from the tsdoc sample https://github.com/microsoft/tsdoc/tree/main/api-demo to only look for comments in declarations.

packages/plugin-scripts/src/manifest/descriptors.ts
rverma3: @actong Yeah. The **isOptional** was added much before the **initialValue** (we can revisit this now). One of the things we need to look out for / think through is that: if a property is optional, it needs an initialValue (at-least with the way the code is structured right now in builder).
 Reasoning behind it:: The property values are usually set one level deep (on clicking the Edit icon). But if the setting is optional, it adds the property directly in the page configuration as soon as the toggle is switched ON(Mustard can then go to edit the value if desired). 
Alternative was that we add the property with 'undefined' , but that was causing complexities with when the toggle was switched off, and the property was removed from the page (which also corresponds to undefined value). So when Mustard would come back to the widget setting, it wasn't clear whether to keep the toggle ON or OFF. So in a nutshell, we decided to make sure there is a value for the setting every time the toggle was switched ON.

Top level **initialValue** was added much later when we identified a scenario where a default value was needed for a required setting as well.

rverma3: I tested with an optional text setting in a custom widget. I also configured an "initialValue" at the top level. At this point:
- the toggle seems to work the way we'd want to (there are minor glitches, but they're more about inconsistently treating an empty string as undefined or not. --- and its not noticeable/causing much issues). 
- the top level "initialValue" was being ignored -- I'll raise a PR in builder to handle that.

But it seems this PR doesn't need to be held up for that, as these changes would work as expected.

One minor thing though -- Please rename **optional** to **isOptional** to match what we've used everywhere so far.

rverma3 (changes requested):
Have added comments inline.

rverma3:
@actong Will you be adding the needsDataSource support in this PR, or are you planning on creating a new PR for it?

actong:
I was going to leave that for a future PR.

rverma3 (approved):

PR #141, opened 2023-07-14
Payment related interfaces and types

ShermanE1:


packages/payment/src/paymentGateway.ts
```diff
@@ -0,0 +1,20 @@
...
+    method?: 'GET' | 'POST';
+    url: string;
+    parameters: Record<string, string>;
+}
+export interface PaymentGateway {
+    prepare(payment: Payment, returnUrl: string): TransferData | PromiseLike<TransferData>;
+    postPaymentResult(
+        payment: Payment,
+        resultData: ParsedQueryString,
+    ): PaymentResult | null | PromiseLike<PaymentResult | null>;
```
actong, line 19: In ML1, PaymentResult was a different type, whereas in Evoke it's all in the same Payment object.  What do you think about having `postPaymentResult` just updating `payment` (or returns the updated payment)?  It seems weird for `Payment` to extend `PaymentResult` (basically saying every payment is a result, which doesn't really make sense).  Also `status` is defined on `PaymentResult`, but really only 'Declined', 'Cancelled', and 'Paid' are really "result" statuses.

packages/payment/src/paymentGateway.ts
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) 2023 System Automation Corporation.
+// This file is licensed under the MIT License.
+
+import { Payment, PaymentResult } from './payment';
+
+export interface ParsedQueryString {
+    [key: string]: undefined | string | string[] | ParsedQueryString | ParsedQueryString[];
+}
+export interface TransferData {
+    method?: 'GET' | 'POST';
```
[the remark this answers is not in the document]
ShermanE1: This should have been required.

packages/payment/src/paymentGateway.ts
ShermanE1: Hmm. I like the idea of having `postPaymentResult` return an updated `payment`, and we can then update the instance that we store to match.

packages/payment/src/payment.ts
```diff
@@ -16,6 +17,7 @@ export interface PaymentItem {
      */
     name: string;
 
+    status: PaymentStatus;
```
actong, line 20: I assume this was meant to go in `Payment`.

packages/payment/src/payment.ts
```diff
@@ -41,7 +90,20 @@ export interface Payment {
     customField9?: string;
     customField10?: string;
 
-    paymentMethod?: PaymentMethod;
+    /**
```
actong, line 93: This is tsdoc syntax, which means this comment is going to be associated as the description for `method`.  Make it a regular comment, not tsdoc, e.g. use `//`.  You can do various things to make it stand out if desired, e.g.
```
// --------------------------
// ...comment goes here....
// --------------------------
```

ShermanE1: Oh okay, I didn't realize that. I'll update this and any other comments that I added that fit into the same category.

actong (approved):

PR #147, opened 2023-07-17
chore: cdr-837-inputProperties export type updated

Usama-Mavric:


packages/context/src/objects/objects.ts
```diff
@@ -33,7 +33,7 @@ export type Action = {
     name: string;
     type: ActionType;
     outputEvent: string;
-    inputProperties?: Property[];
+    inputProperties?: Record<string, unknown>[];
```
actong, line 36: I assume we wanted a more specific type?  @rverma3

Usama-Mavric: Do we want a type incorporating all the different properties from every FormIO field type? 
This example below doesn't have properties that the "Select" field will use i.e. missing fields like data.values etc.  
If that's the case then we can merge all the field props from all formIo fields into one single type.  
export type FormIoInputType= {
    html: string,
    label: string,
    customClass?: string,
    refreshOnChange: boolean,
    hidden: boolean,
    modalEdit: boolean,
    conditional: unknown,
    type: string,
    key: string,
    input: boolean,
    placeholder?: string,
    prefix?: string,
    suffix?: string,
    multiple: boolean,
    defaultValue: null,
    protected: boolean,
    unique: boolean,
    persistent:boolean,
    clearOnHide:boolean,
    refreshOn?: string,
    redrawOn?: string,
    tableView: boolean,
    dataGridLabel: boolean,
    labelPosition: string,
    description?: string,
    errorLabel?: string,
    tooltip?: string,
    hideLabel: boolean,
    tabindex?: string,
    disabled: boolean,
    autofocus: boolean,
    dbIndex: boolean,
    customDefaultValue?: string,
    calculateValue?: string,
    calculateServer: boolean,
    widget: null,
    attributes?: unknown,
    validateOn: string,
    validate: unknown
    overlay: unknown
    allowCalculateOverride: boolean,
    encrypted: boolean,
    showCharCount: boolean,
    showWordCount: boolean,
    properties?: unknown,
    allowMultipleMasks: boolean,
    addons: any [],
    id: string,
}

actong: The idea is for custom widget developers to know how to populate action input.  I don't know if all of that is necessary, but at least some subset of it (certainly important ones like `type`, `key`, `label`).  Should probably work with @rverma3 on what should be included.

packages/context/src/objects/objects.ts
```diff
@@ -43,6 +43,37 @@ export type ObjectInstance = {
     [key: string]: unknown;
 };
 
+export type FormioInputTypes = {
+    label: string;
+    type: string;
+    key: string;
+    initialValue?: any;
```
actong, line 50: You can use `unknown` instead.

packages/context/src/objects/objects.ts
```diff
@@ -43,6 +43,37 @@ export type ObjectInstance = {
...
+    property?: Property;
```
actong, line 51: Out of curiosity, what is this used for?  What sets this in Builder?

packages/context/src/objects/objects.ts
```diff
@@ -43,6 +43,37 @@ export type ObjectInstance = {
...
+    input?: boolean;
```
actong, line 52: Same question here, how is this set in Builder and what is it used for?

packages/context/src/objects/objects.ts
```diff
@@ -43,6 +43,37 @@ export type ObjectInstance = {
     [key: string]: unknown;
 };
 
+export type FormioInputTypes = {
```
actong, line 46: Don't think this needs to be plural, but beyond that, I think it can just be `FormioInput` or `FormIOInput`, don't really need to say it's a type, e.g. just like we don't say `ObjectInstanceType`.

packages/context/src/objects/objects.ts
```diff
@@ -43,6 +43,37 @@ export type ObjectInstance = {
...
+    data?: {
+        resource?: string;
+        values?: [];
+    };
+    dataSrc?: string;
+    validate?: {
+        required: boolean;
+        operator: any;
+    };
+    components: FormioInputTypes[];
```
actong, line 73: Is this for the Section layout field type?  Maybe it's required for sections, but in the grand scheme, I'd think this should be marked optional, or we need more intricate typings to show that it's only required for sections.

Can you also add a tsdoc here to explain what this is?

packages/context/src/objects/objects.ts
```diff
@@ -43,6 +43,37 @@ export type ObjectInstance = {
...
+    input?: boolean;
+    placeholder?: string;
+    description?: string;
+    tooltip?: string;
+    prefix?: string;
+    suffix?: string;
+    showCharCount?: boolean;
+    readOnly?: boolean;
+    isMultiLineText?: boolean;
+    widget?: string;
```
actong, line 61: What are `widget`, `tableView`, `verticalLayout`, `data`, and `dataSrc` for?

packages/context/src/objects/objects.ts
```diff
@@ -43,6 +43,37 @@ export type ObjectInstance = {
...
+    tableView?: boolean;
+    verticalLayout?: boolean;
+    data?: {
+        resource?: string;
+        values?: [];
+    };
+    dataSrc?: string;
+    validate?: {
+        required: boolean;
+        operator: any;
```
actong, line 71: Does this correspond to the "Must match regular expression rules" validation setting in Builder?  If so, I assume we know the type, it shouldn't be `any`.  Also, I don't see where you specify the validation regular expressions that go with it.

packages/context/src/objects/objects.ts
Usama-Mavric: In the builder, whenever we drag and drop a field in the form this `property` automatically gets added to the JSON via the code.

packages/context/src/objects/objects.ts
Usama-Mavric: This is getting set via code if the field is calculated one then the input will be false otherwise will be true.

packages/context/src/objects/objects.ts
Usama-Mavric: The `data` contains the values for the `select` options. The `dataSrc` is referencing where to get the values from i.e. `dataSrc: "values"`.
The `verticalLayout` prop will be used for section layout if we want a vertical layout or not. It's hard-coded. There's no way to set the `verticalLayout` and `tableView` in the builder and I believe this could be removed.

packages/context/src/objects/objects.ts
Usama-Mavric: Done.

packages/context/src/objects/objects.ts
Usama-Mavric: Done.

packages/context/src/objects/objects.ts
Usama-Mavric: Done.

packages/context/src/objects/objects.ts
Usama-Mavric: Added the regexes [].

packages/context/src/objects/objects.ts
```diff
@@ -43,6 +43,40 @@ export type ObjectInstance = {
...
+    placeholder?: string;
+    description?: string;
+    tooltip?: string;
+    prefix?: string;
+    suffix?: string;
+    showCharCount?: boolean;
+    readOnly?: boolean;
+    isMultiLineText?: boolean;
+    data?: {
+        resource?: string;
```
actong, line 67: What is `resource`?

packages/context/src/objects/objects.ts
Usama-Mavric: This will render the value inside the field and the field will not be readOnly. I'm not sure whether to add this to the type definition but the logic in the code is like this `input: !property.formula,`

packages/context/src/objects/objects.ts
Usama-Mavric: I believe this could be retrieved from the `objectProperties` if we don't set it in the builder. This can be removed.

packages/context/src/objects/objects.ts
Usama-Mavric: I tried changing the `dataSrc` to some random string but the formio still getting the values from `data.values`. This can also be removed.

packages/context/src/objects/objects.ts
Usama-Mavric: This is being added by FormIO, removing it will not have any effect on the form.

packages/context/src/objects/objects.ts
```diff
@@ -43,11 +43,16 @@ export type ObjectInstance = {
     [key: string]: unknown;
 };
 
-export type FormioInputTypes = {
+/**
+ * @typedef {object} FormioInput - creates a new type named 'FormioInput'
```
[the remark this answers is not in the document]
Usama-Mavric: Removed.

packages/context/src/objects/objects.ts
Usama-Mavric: Removed.

packages/context/src/objects/objects.ts
```diff
@@ -43,6 +43,51 @@ export type ObjectInstance = {
...
+    readOnly?: boolean;
+    isMultiLineText?: boolean;
+    /**
+     * An array of values required for select options.
+     */
+    data?: {
+        values?: SelectOption[];
+    };
+    validate?: {
+        required: boolean;
```
actong, line 80: Please double check if all these fields (`required`, `operator`, `regexes`) are required.  I would assume they are not.

packages/context/src/objects/objects.ts
```diff
@@ -43,6 +43,51 @@ export type ObjectInstance = {
...
+    placeholder?: string;
+    description?: string;
+    tooltip?: string;
+    prefix?: string;
+    suffix?: string;
+    showCharCount?: boolean;
+    readOnly?: boolean;
+    isMultiLineText?: boolean;
+    /**
+     * An array of values required for select options.
```
actong, line 74: This should go with `data.values`, not `data`.

packages/context/src/objects/objects.ts
actong: @rverma3 Can you comment on what `input` is for?

packages/context/src/objects/objects.ts
Usama-Mavric: You are right. These are optional I checked. Updated.

packages/context/src/objects/objects.ts
Usama-Mavric: Done.

packages/context/src/objects/objects.ts
Usama-Mavric: Input prop removed.

packages/context/src/objects/objects.ts
```diff
@@ -43,11 +43,16 @@ export type ObjectInstance = {
     [key: string]: unknown;
 };
 
-export type FormioInputTypes = {
+/**
+ * @typedef {object} FormioInput - creates a new type named 'FormioInput'
+ * This will be used if a form field needs to be configured for custom widget
+ * @property { FormioInput } components - a property of FormioInput type groups different sets of fields together into separate sections.
+ */
+export type FormioInput = {
```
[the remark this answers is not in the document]
rverma3: Hmm.. no i think we should not be making it obvious.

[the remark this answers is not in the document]
Usama-Mavric: Re-named to ActionInput.

actong (approved):

PR #167, opened 2023-08-08
chore: add slug property to page type

ShermanE1:


packages/context/src/app/AppProvider.tsx
```diff
@@ -20,6 +20,7 @@ export type App = {
 export type Page = {
     id: string;
     name: string;
+    slug: string;
```
ShermanE1, line 23: Thought I made this optional, but I guess I never saved the update. Will update.

actong (changes requested):
Can you rebase and update the commit messages?  Backwards compatible interface changes are a `feat`, not a `chore`.

ShermanE1:
Okay, I'll rebase. I was thinking that a `feat` is for breaking changes and `chore` is for backward compatible ones.

PR #283, opened 2024-01-10
feat: add notification provider

mtruongsa:
Set session id to api request headers

### DIFF
--- a/packages/context/README.md
+++ b/packages/context/README.md
@@ -199,7 +199,7 @@ absolute URL.
 
 ### SignalR Connection
 
-@depreacated
+Deprecated
 
 -   [useSignalRConnection](#usesignalrconnection)
     -   [documentChanges](#documentchangessubscribeobjectidinstanceid-data-documentchange)
@@ -272,6 +272,7 @@ The data provided to the callback will be an array of instance IDs that were upd
 Unsubscribe to the specified object instance changes.
 
 Callback function is optional.
+If callback function is not defined, nothing will not be invoked when a connection is closed.
 If callback function is defined, you must pass the exact same Function instance as was previously passed to `instanceChanges.subscribe`.
 Passing a different instance (even if the function body is the same) will not remove the subscription.
 
@@ -287,7 +288,7 @@ instanceChanges.subscribe('myObjectId', callback);
 instanceChanges.unsubscribe('myObjectId', callback);
 ```
 
-### Notification Connection
+### Notification
 
 -   [useNofitication](#usenotification)
     -   [documentChanges](#documentchangessubscribeobjectidinstanceid-data-documentchange)
@@ -295,11 +296,11 @@ instanceChanges.unsubscribe('myObjectId', callback);
 
 #### `useNofitication()`
 
-Hook used to obtain an instanceChanges instance of `Notification` and a documentChanges instance of `Notification`.
+Hook used to obtain an instanceChanges instance and a documentChanges instance.
 
 ##### `documentChanges.subscribe(objectId, instanceId, (data: DocumentChange[]]) => {})`
 
-Subscribe to the specified object instance document changes.
+Subscribe to the specified object instance changes.
 
 ```javascript
 const { documentChanges } = useNotification();
@@ -321,19 +322,20 @@ following data:
 -   `type`
     -   The type of update. Possible values are `BlobCreated`, `BlobDeleted`, and `BlobMetadataUpdated`.
 
-##### `documentChanges.unsubscribe(objectId, instanceId, (data: DocumentChange[]) => {})`
+##### `documentChanges.unsubscribe(objectId, instanceId, (changes: DocumentChange[]) => {})`
 
-Unsubscribe to the specified object instance document changes.
+Unsubscribe to the specified object instance changes.
 
 Callback function is optional.
+If callback function is not defined, nothing will not be invoked when a connection is closed.
 If callback function is defined, you must pass the exact same Function instance as was previously passed to `documentChanges.subscribe`.
 Passing a different instance (even if the function body is the same) will not remove the subscription.
 
 ```javascript
 const { documentChanges } = useNotification();
 
-const callback = (data: DocumentChange[]) => {
-    console.log(data);
+const callback = (changes: DocumentChange[]) => {
+    console.log(changes);
 };
 
 documentChanges.subscribe('myObjectId' , 'myInstanceId', callback);
@@ -341,23 +343,29 @@ documentChanges.subscribe('myObjectId' , 'myInstanceId', callback);
 documentChanges.unsubscribe('myObjectId', 'myInstanceId', callback);
 ```
 
-##### `instanceChanges.subscribe(objectId, undefined, (instanceIds: InstanceChange[]) => {})`
+##### `instanceChanges.subscribe(objectId, (changes: InstanceChange[]) => {})`
 
-Subscribe to the specified object instance changes.
+Subscribe to the specified object changes.
 
 ```javascript
 const { instanceChanges } = useNotification();
 
-instanceChanges.subscribe('myObjectId', undefined, (instanceIds) => {
-    console.log(instanceIds);
+instanceChanges.subscribe('myObjectId', (changes) => {
+    console.log(changes);
 });
 ```
 
-The data provided to the callback will be an array of instance IDs that were updated.
+The data provided to the callback will be an array of `InstanceChange` which contains the
+following data:
 
-##### `instanceChanges.unsubscribe(objectId, undefined, (instanceIds: InstanceChange[]) => {})`
+-   `objectId`
+    -   Object describing the instance associated with the updated document.
+-   `instanceId`
+    -   Instance that the updated document is associated with.
 
-Unsubscribe to the specified object instance changes.
+##### `instanceChanges.unsubscribe(objectId, (changes: InstanceChange[]) => {})`
+
+Unsubscribe to the specified object changes.
 
 Callback function is optional.
 If callback function is defined, you must pass the exact same Function instance as was previously passed to `instanceChanges.subscribe`.
@@ -366,13 +374,13 @@ Passing a different instance (even if the function body is the same) will not re
 ```javascript
 const { instanceChanges } = useNotification();
 
-const callback = (instanceIds: InstanceChange[]) => {
-    console.log(instanceIds);
+const callback = (changes: InstanceChange[]) => {
+    console.log(changes);
 };
 
-instanceChanges.subscribe('myObjectId', undefined, callback);
+instanceChanges.subscribe('myObjectId', callback);
 
-instanceChanges.unsubscribe('myObjectId', undefined, callback);
+instanceChanges.unsubscribe('myObjectId', callback);
 ```
 
 ## License
--- a/packages/context/README.md
+++ b/packages/context/README.md
@@ -272,7 +272,6 @@ The data provided to the callback will be an array of instance IDs that were upd
 Unsubscribe to the specified object instance changes.
 
 Callback function is optional.
-If callback function is not defined, nothing will not be invoked when a connection is closed.
 If callback function is defined, you must pass the exact same Function instance as was previously passed to `instanceChanges.subscribe`.
 Passing a different instance (even if the function body is the same) will not remove the subscription.
 
@@ -327,7 +326,7 @@ following data:
 Unsubscribe to the specified object instance changes.
 
 Callback function is optional.
-If callback function is not defined, nothing will not be invoked when a connection is closed.
+If callback function is not defined, all subscriptions will be removed.
 If callback function is defined, you must pass the exact same Function instance as was previously passed to `documentChanges.subscribe`.
 Passing a different instance (even if the function body is the same) will not remove the subscription.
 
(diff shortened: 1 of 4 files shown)


packages/context/src/notification/NotificationProvider.tsx
```diff
@@ -0,0 +1,182 @@
...
+                            if (instancesConnectionInfo.accessToken) {
+                                return instancesConnectionInfo.accessToken;
+                            } else {
+                                return getConnection();
+                            }
+                        },
+                    };
+
+                    const connection = new HubConnectionBuilder()
+                        .withUrl(instancesConnectionInfo.url, options as unknown as HttpTransportType)
```
actong, line 62: Why is `as unknown` needed?  I think this is pointing out that there's a problem with `accessTokenFactory`.

packages/context/src/notification/NotificationProvider.tsx
```diff
@@ -0,0 +1,182 @@
...
+                const instancesConnectionInfo = await getConnectionInfo('instanceChanges');
+                const documentsConnectionInfo = await getConnectionInfo('documentChanges');
+
+                if (instancesConnectionInfo) {
+                    const options = {
+                        accessTokenFactory: async () => {
+                            if (instancesConnectionInfo.accessToken) {
+                                return instancesConnectionInfo.accessToken;
+                            } else {
+                                return getConnection();
```
actong, line 56: I assume `accessTokenFactory` is supposed to return a token, but `getConnection` doesn't return anything.

packages/context/src/notification/NotificationProvider.tsx
```diff
@@ -0,0 +1,182 @@
...
+                    const connection = new HubConnectionBuilder()
+                        .withUrl(documentsConnectionInfo.url, options as unknown as HttpTransportType)
+                        .configureLogging(LogLevel.Error)
+                        .withAutomaticReconnect()
+                        .build();
+
+                    setDocumentsNotification(connection);
+                }
+                // eslint-disable-next-line no-empty
+            } catch (err) {}
```
actong, line 90: Should at least log the error

packages/context/src/notification/NotificationProvider.tsx
```diff
@@ -0,0 +1,182 @@
...
+};
+
+export type DocumentChange = {
+    objectId: string;
+    instanceId: string;
+    documentId: string;
+    type: string;
+};
+
+export type InstanceChange = string;
```
actong, line 23: Can we also do something like `DocumentChange` and make this e.g.
```
{
  objectId: string;
  instanceId: string;
}
```

packages/context/src/notification/NotificationProvider.tsx
```diff
@@ -0,0 +1,182 @@
...
+import { HttpTransportType, HubConnection, HubConnectionBuilder, LogLevel } from '@microsoft/signalr/dist/esm/index.js';
+import { useApiServices } from '../api/index.js';
+
+export type NotificationConnectionInfo = {
+    url: string;
+    accessToken: string;
+};
+
+export type Subscription<T> = {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
actong, line 11: Why does the callback return `any`?  Do we do anything with that returned value?

packages/context/src/notification/NotificationProvider.tsx
```diff
@@ -0,0 +1,182 @@
...
+    subscribe: (topic: string, callback: (...args: T[]) => any) => void;
```
actong, line 12: I think we need something more descriptive than `topic`.  E.g. if the intent is to subscribe to document changes for a specific object instance, then explicitly asking the caller to provide `objectId` and `instanceId` would make for an easier to use api.  Otherwise they have to know how to construct these topic names (btw what are examples of current topic names?).

packages/context/src/api/apiServices.ts
```diff
@@ -6,9 +6,12 @@ import { useMemo } from 'react';
 import { AuthenticationContext, useAuthenticationContext } from '../authentication/AuthenticationContextProvider.js';
 import { useApiBaseUrl } from './ApiBaseUrlProvider.js';
 import { Callback } from './callback.js';
+import { randomUUID } from 'crypto';
 
 export type Data = Record<string, unknown> | FormData;
 
+const sessionId = randomUUID();
```
actong, line 13: This needs to be integrated with `NotificationProvider`.  When a widget subscribes to receive notifications, the `subscribe` needs to know the session id.

actong (commented):
Can you update the README with documentation about how to use the new API?  Also should mark the old API as deprecated and point users at using the new API instead.

packages/context/src/notification/NotificationProvider.tsx
mtruongsa: we are not doing anything. this can return void

packages/context/src/api/apiServices.ts
mtruongsa: The current implementation uses the X-Forwarded-Host to restrict who can see messages. That means everyone in the same environment can see it. Now, we uses X-Session-Id to restrict who can see messages.

packages/context/src/notification/NotificationProvider.tsx
mtruongsa: For document changes, we construct it as `${objectId}/${instanceId}`

packages/context/src/notification/NotificationProvider.tsx
actong: From an api standpoint, yeah, there's no reason to tie ourselves to signalr, but I think it's fine for diagnostic logging.

packages/context/src/notification/NotificationProvider.tsx
actong: Does it do anything with that return value?

mtruongsa: behind the scene, it is used the return value to set the complete message (probably logging to their system).

actong: What do you mean by "to set the complete message"?

mtruongsa: their server is expecting a response so they use the callback returned value to construct a completion message.

actong: But is that only if the server requests a response?  I.e. do we have to do something on the server side for it to expect the response, otherwise it's just ignored?

packages/context/README.md
```diff
@@ -199,6 +199,8 @@ absolute URL.
 
 ### SignalR Connection
 
+@depreacated
```
actong, line 202: Typo `deprecated`.

Also, is this some markdown syntax, or why the reason for `@`?

packages/context/README.md
```diff
@@ -285,6 +287,94 @@ instanceChanges.subscribe('myObjectId', callback);
 instanceChanges.unsubscribe('myObjectId', callback);
 ```
 
+### Notification Connection
```
actong, line 290: "Connection" is unnecessary I think, this is just about getting change notifications.

packages/context/README.md
```diff
@@ -285,6 +287,94 @@ instanceChanges.subscribe('myObjectId', callback);
...
+
+-   [useNofitication](#usenotification)
+    -   [documentChanges](#documentchangessubscribeobjectidinstanceid-data-documentchange)
+    -   [instanceChanges](#instancechangessubscribeobjectid-instanceids-instancechange)
+
+#### `useNofitication()`
+
+Hook used to obtain an instanceChanges instance of `Notification` and a documentChanges instance of `Notification`.
```
actong, line 298: This wording seems confusing.  What is the `Notification` this is referering to?

packages/context/README.md
```diff
@@ -285,6 +287,94 @@ instanceChanges.subscribe('myObjectId', callback);
...
+-   `documentId`
+    -   Document that was updated.
+-   `type`
+    -   The type of update. Possible values are `BlobCreated`, `BlobDeleted`, and `BlobMetadataUpdated`.
+
+##### `documentChanges.unsubscribe(objectId, instanceId, (data: DocumentChange[]) => {})`
+
+Unsubscribe to the specified object instance document changes.
+
+Callback function is optional.
```
actong, line 328: Should describe what happens if you don't specify the callback.

packages/context/README.md
```diff
@@ -285,6 +287,94 @@ instanceChanges.subscribe('myObjectId', callback);
...
+const callback = (data: DocumentChange[]) => {
+    console.log(data);
+};
+
+documentChanges.subscribe('myObjectId' , 'myInstanceId', callback);
+
+documentChanges.unsubscribe('myObjectId', 'myInstanceId', callback);
+```
+
+##### `instanceChanges.subscribe(objectId, undefined, (instanceIds: InstanceChange[]) => {})`
```
actong, line 344: `instanceChanges` only needs `objectId`, there's no reason to have a second argument that's always ignored.

Also, just call the callback parameter something like `changes`.  They're not just instance ids.

packages/context/README.md
```diff
@@ -285,6 +287,94 @@ instanceChanges.subscribe('myObjectId', callback);
...
+
+Subscribe to the specified object instance changes.
```
actong, line 346: You only specify the object, caller is subscribing to instance changes of the specified object.

packages/context/README.md
```diff
@@ -285,6 +287,94 @@ instanceChanges.subscribe('myObjectId', callback);
...
+
+```javascript
+const { instanceChanges } = useNotification();
+
+instanceChanges.subscribe('myObjectId', undefined, (instanceIds) => {
```
actong, line 351: Don't use `instanceIds`, that's misleading.

packages/context/src/notification/NotificationProvider.tsx
mtruongsa: https://github.com/dotnet/aspnetcore/blob/f545e80531c098e7c1611eca861b0f080fa14a3c/src/SignalR/clients/ts/signalr/src/HubConnection.ts#L756C1-L756C1

This is their implementation. I don't think we need to do anything.

packages/context/README.md
```diff
@@ -285,6 +287,94 @@ instanceChanges.subscribe('myObjectId', callback);
...
+    console.log(instanceIds);
+});
+```
+
+The data provided to the callback will be an array of instance IDs that were updated.
```
actong, line 356: Update to reflect actual returned data.

packages/context/src/notification/NotificationProvider.tsx
```diff
@@ -48,18 +51,14 @@ function NotificationProvider({ children }: { children: React.ReactNode }) {
...
-                    const options = {
-                        accessTokenFactory: async () => {
-                            if (instancesConnectionInfo.accessToken) {
-                                return instancesConnectionInfo.accessToken;
-                            } else {
-                                return getConnection();
-                            }
+                    const options: IHttpConnectionOptions = {
+                        accessTokenFactory: () => {
+                            return instancesConnectionInfo.accessToken ?? '';
```
actong, line 56: Do you know how long this token is valid for?

packages/context/src/notification/NotificationProvider.tsx
actong:
```
// Server expects a response
const expectsResponse = invocationMessage.invocationId ? true : false;
```

This suggests to me they don't always expect a response.  In any case, we certainly don't expect a response, so for users of the evoke sdk, this should just be a `void` return.  If signalr wants a response, that's an implementation detail that would be handled by the sdk.

mtruongsa: got it.

packages/context/src/notification/NotificationProvider.tsx
mtruongsa: I believe it is 1 hour. We will need to refresh it once it is expired.

mtruongsa: https://github.com/dotnet/AspNetCore.Docs/issues/13144 I found this interesting. They don't check if access token expires.

actong: What happens after 1 hour?  Do we stop receiving notifications?

packages/context/README.md
mtruongsa: No it is not

packages/context/README.md
```diff
@@ -321,43 +322,50 @@ following data:
...
     -   The type of update. Possible values are `BlobCreated`, `BlobDeleted`, and `BlobMetadataUpdated`.
 
-##### `documentChanges.unsubscribe(objectId, instanceId, (data: DocumentChange[]) => {})`
+##### `documentChanges.unsubscribe(objectId, instanceId, (changes: DocumentChange[]) => {})`
 
-Unsubscribe to the specified object instance document changes.
+Unsubscribe to the specified object instance changes.
 
 Callback function is optional.
+If callback function is not defined, nothing will not be invoked when a connection is closed.
```
actong, line 330: You've got a double negative "nothing will not be invoked", and I'm not exactly sure what this is saying, but it sounds like this is saying it will be a no-op.  I don't think that's the case, I'm not sure, but from the SignalR documentation it sounded like it would unsubscribe everything.

packages/context/README.md
```diff
@@ -321,43 +322,50 @@ following data:
...
+    -   Object describing the instance associated with the updated document.
+-   `instanceId`
+    -   Instance that the updated document is associated with.
 
-Unsubscribe to the specified object instance changes.
+##### `instanceChanges.unsubscribe(objectId, (changes: InstanceChange[]) => {})`
+
+Unsubscribe to the specified object changes.
 
 Callback function is optional.
```
actong, line 370: Describe behavior if callback not provided.

PR #293, opened 2024-01-15
feat: payment failsafe function

tlachman:
Add `receivePaymentNotification` function. This function is called when the notification endpoint is called by the payment processor.

### DIFF
--- a/packages/payment/package.json
+++ b/packages/payment/package.json
@@ -30,8 +30,5 @@
     "devDependencies": {
         "commit-and-tag-version": "^12.1.0",
         "typescript": "^5.3.2"
-    },
-    "dependencies": {
-        "@loopback/rest": "^13.1.5"
     }
 }
--- a/packages/payment/src/paymentGateway.ts
+++ b/packages/payment/src/paymentGateway.ts
@@ -1,7 +1,7 @@
 // Copyright (c) 2023 System Automation Corporation.
 // This file is licensed under the MIT License.
 
-import type { Request, Response } from '@loopback/rest';
+import { IncomingMessage, ServerResponse } from 'http';
 import { Payment } from './payment';
 
 export interface ParsedQueryString {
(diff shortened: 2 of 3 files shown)


packages/payment/package.json
```diff
@@ -30,5 +30,8 @@
     "devDependencies": {
         "commit-and-tag-version": "^12.1.0",
         "typescript": "^5.3.2"
+    },
+    "dependencies": {
+        "@loopback/rest": "^13.1.5"
```
actong, line 35: Maybe this should be a peer dependency.  The request/response are being provided by the Evoke runtime and will be for whatever version Evoke is using, not the plugin's version.

Also, I believe all Loopback does is re-export these from `express`, so maybe get them from `express` instead of adding Loopback as a dependency.  And `express` request/response extend node's `http.IncomingMessage` and `http.ServerResponse`, so we could maybe even just use those and not need any dependency or worry about if e.g. we upgrade Loopback and it starts using a new major version of `express`.

packages/payment/src/paymentGateway.ts
```diff
@@ -16,4 +17,10 @@ export interface TransferData {
 export interface PaymentGateway {
     prepare(payment: Payment, returnUrl: string): TransferData | PromiseLike<TransferData>;
     postPaymentResult(payment: Payment, resultData: ParsedQueryString): Payment | null | PromiseLike<Payment | null>;
+    receivePaymentNotification(
```
actong, line 20: Adding this would technically be a breaking change and would require a major version bump since existing implementations don't have this.  Alternatively we can make this optional.  Evoke itself will need to treat it as optional anyway since there may be implementations without this.

packages/payment/src/paymentGateway.ts
```diff
@@ -16,4 +17,10 @@ export interface TransferData {
...
+        request: Request,
+        response: Response,
+    ): void | Payment | null | PromiseLike<Payment | null>;
```
actong, line 23: Does this need `void`?

actong (approved):

PR #294, opened 2024-01-18
feat: export failsafe request and response params from express

tlachman:


packages/payment/src/paymentGateway.ts
```diff
@@ -14,9 +14,6 @@ export interface TransferData {
     parameters: Record<string, string>;
 }
 
-export type Request = IncomingMessage;
-export type Response = ServerResponse;
```
actong, line 18: Do we still want to re-export these?

actong (approved):

PR #314, opened 2024-02-08
feat: add loginHint to AuthenticationRequest type

tlachman:


packages/context/src/authentication/AuthenticationContextProvider.tsx
```diff
@@ -26,7 +26,7 @@ export type AuthenticationContextProviderProps = {
     children?: ReactNode;
 };
 
-export type AuthenticationRequest = Pick<RedirectRequest, 'scopes' | 'extraQueryParameters' | 'state'>;
+export type AuthenticationRequest = Pick<RedirectRequest, 'scopes' | 'extraQueryParameters' | 'state' | 'loginHint'>;
```
actong, line 29: Is `loginHint` something that has to be provided by the enclosing code, i.e. the context library can't figure it out itself?  I would have thought the point of passing `account` in `acquireToken*` was to set the login hint.

PR #330, opened 2024-02-21
feat: update object types to match model

ShermanE1:


packages/context/src/objects/objects.ts
```diff
@@ -34,25 +59,62 @@ export type PropertyType =
...
+    errorMessage?: string;
+    to?: string;
+    from?: string;
+};
+
+export type CriteriaValidation = {
+    criteria?: Record<string, unknown>;
+};
+
+export type PropertyValidation = RegexValidation | NumericValidation | DateValidation | CriteriaValidation;
```
actong, line 78: `RegexValidation` doesn't match what the services use.

packages/context/src/objects/objects.ts
```diff
@@ -72,38 +134,164 @@ export type SelectOption = {
...
+export type Section = {
+    label: string;
+    entries?: FormEntry[];
+};
+
+export type Sections = {
+    type: 'sections';
+    label?: string;
+    sections: Section[];
+    visibility?: VisibilityConfiguration | string;
```
actong, line 198: I see this is how it is in the services, but why is there `| string` here?

packages/context/src/objects/objects.ts
```diff
@@ -72,38 +134,164 @@ export type SelectOption = {
     value: string;
 };
 
+export type VisibilityConfiguration = {
+    operator?: 'any' | 'all';
```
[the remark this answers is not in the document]
ShermanE1: We default to `all` in the front end.

packages/context/src/objects/objects.ts
ShermanE1: FormIO supports more complex conditions in JSON format - which is saved as a string.

packages/context/src/objects/objects.ts
```diff
@@ -124,11 +132,6 @@ export type ObjectInstance = {
     [key: string]: unknown;
 };
 
-export type RegexValidation = {
```
actong, line 65: If you remove this exported type, technically it would make this a breaking change.  Any reason for removing it?

packages/context/src/objects/objects.ts
actong: If it is a JSON string, why wouldn't it be saved as an object instead?  Does FormIO actually expect a JSON string that it then parses into some object?

packages/context/src/objects/objects.ts
```diff
@@ -62,48 +132,172 @@ export type ObjectInstance = {
...
+        maxTime?: string;
+        min?: number;
+        max?: number;
+        customMessage?: string;
     };
     /**
      * An array of sub-components to be rendered inside sections.
      */
-    components?: ActionInput[];
-    [key: string]: unknown;
```
actong, line 106: Removing this is technically a breaking change.  If we want to do it, I'm fine with it, but it should be marked as such in the commits so the versions are updated appropriately.

packages/context/src/objects/objects.ts
ShermanE1: No particular reason. I'll add this back in.

packages/context/src/objects/objects.ts
ShermanE1: Yeah, they expect a string. I'm guessing they parse it behind the scenes.

We could technically save it as an object in our database, and "stringify" it when converting to the FormIO components.

packages/context/src/objects/objects.ts
ShermanE1: Got it. I'll rebase to update the commit.

packages/context/src/objects/objects.ts
actong: It's fine.  In any case I wasn't expecting this to change that anyway since it requires services to change, was just curious why it was a string.  But if that's what FormIO expects, I think we can just leave it.

actong (approved):

actong (approved):

PR #335, opened 2024-02-26
feat: add missing properties to action type

ShermanE1:


mtruongsa (approved):

packages/context/src/objects/objects.ts
```diff
@@ -216,34 +220,39 @@ export type Form = {
     entries?: FormEntry[];
 };
 
+export type ActionInputType =
+    | 'button'
```
actong, line 224: Is this supposed to be lower cased?  Also, what does this correspond to in a form?  Do we have buttons in forms?

ShermanE1: This corresponds to the submit/cancel buttons on a form. We append them to the form on the FE, so they don't ever get saved to the DB. 

We filter out these buttons from the form definition in some places on the FE once they are added in, so I added it here as we are using this type definition.

ShermanE1: Though if it is more accurate from a data perspective, I can append/override this type in the FE.

actong: What could do the button do?  Is it just the same as the submit/cancel buttons we have, or can they pretty much do anything with it?

ShermanE1: It seems like pretty much anything. FormIO expects an `onClick` handler.

actong (approved):

PR #338, opened 2024-02-29
feat: update action input type to include many-to-many property

ShermanE1:


actong:
@ShermanE1 Is this still something you want merged?

PR #356, opened 2024-03-21
feat: update input parameter model

mtruongsa:


ShermanE1 (approved):

packages/context/src/objects/objects.ts
```diff
@@ -116,10 +116,14 @@ export type InputStringValidation = StringValidation & {
 
 export type InputParameter = {
     id: string;
+    name: string;
```
actong, line 119: Inputs don't have `name`

actong: Nevermind, I see it just got added.

PR #385, opened 2024-04-11
feat: add function to get default page slug

bradwray:


packages/context/README.md
```diff
@@ -160,7 +160,16 @@ Returns a function that can be used to navigate to another page. The returned fu
 
 #### `useApp()`
 
-Returns the currently loaded Evoke app.
+Returns an object with the following properties:
```
actong, line 163: The previous description is still accurate, it is still returning the Evoke app.  I think the new description ("an object") makes it more nebulously what it returns, and also it's not accurate; it does not have an `app` property.

packages/context/README.md
```diff
@@ -160,7 +160,16 @@ Returns a function that can be used to navigate to another page. The returned fu
...
+
+-   `app`: The currently loaded Evoke app.
+-   `findDefaultPageSlugFor`: An asynchronous function that takes an `objectId` as a parameter and returns the default page slug for that object, if no default page slug is found and the object is a subtype, the page slug of the first ancestor with a default page will be used. It returns `undefined` if no default page slug is found for any ancestor type.
+
+Example usage:
+
+````javascript
```
actong, line 170: Is there some significance to having 4 backticks instead of 3?

Also, this code block is not terminated.

packages/context/src/app/AppProvider.tsx
```diff
@@ -62,8 +78,57 @@ export type AppProviderProps = {
...
 function AppProvider(props: AppProviderProps) {
     const { app, children } = props;
-
-    return <AppContext.Provider value={app}>{children}</AppContext.Provider>;
+    const apiServices = useApiServices();
+
+    const appWithFunctions: AppWithFunctions = {
+        ...app,
+        /**
+         * Looks up the default page slug for a given object or its nearest type ancestor.
```
actong, line 86: Comment should be part of the type definition, i.e. it should show up in the generated `*.d.ts` file.

packages/context/src/app/AppProvider.tsx
```diff
@@ -62,8 +78,57 @@ export type AppProviderProps = {
...
+                while (currentObjectId !== undefined) {
+                    if (app.defaultPages?.[currentObjectId]) {
+                        defaultPageId = app.defaultPages[currentObjectId];
+                        break;
+                    }
+
+                    let effectiveObject: Obj | undefined;
+                    try {
+                        effectiveObject = await apiServices.get<Obj>(`data/objects/${currentObjectId}/effective`, {
+                            params: { filter: { fields: ['baseObject'] }, sanitizedVersion: true },
```
actong, line 105: Don't really need `sanitizedVersion`, this is only retrieving `baseObject`.  `sanitizedVersion` only affects what properties and actions are returned.

packages/context/src/app/AppProvider.tsx
```diff
@@ -62,8 +78,57 @@ export type AppProviderProps = {
...
+                        });
+                    } catch (error) {
+                        console.error(error);
```
actong, line 108: What's the reason for swallowing the error?

packages/context/src/app/AppProvider.tsx
```diff
@@ -62,8 +78,57 @@ export type AppProviderProps = {
...
+        /**
+         * Looks up the default page slug for a given object or its nearest type ancestor.
+         *
+         * @param {string} objectId - The ID of the object to start the search from.
+         * @returns {Promise<string | undefined>} The default page slug, or `undefined` if no default page is found.
+         */
+        findDefaultPageSlugFor: useCallback(
+            async (objectId: string) => {
+                let defaultPageId: string | undefined;
+                const objectHierarchy = [objectId];
```
actong, line 94: Doesn't seem like `objectHierarchy` serves any purpose.

packages/context/src/app/AppProvider.tsx
```diff
@@ -62,8 +78,57 @@ export type AppProviderProps = {
...
+                        console.error(error);
+                    }
+                    if (effectiveObject?.baseObject?.objectId) {
+                        objectHierarchy.push(effectiveObject?.baseObject?.objectId);
+                    }
+                    currentObjectId = effectiveObject?.baseObject?.objectId ?? undefined;
+                }
+
+                let defaultPage: Page | undefined;
+                if (defaultPageId && typeof defaultPageId === 'string') {
```
actong, line 117: When would `defaultPageId` not be a string (other than when it's undefined)?

packages/context/src/app/AppProvider.tsx
```diff
@@ -62,8 +78,57 @@ export type AppProviderProps = {
...
+                    const pageId = defaultPageId.includes('/')
+                        ? defaultPageId.split('/').slice(2).join('/')
```
actong, line 119: So I guess we're prepending the app slug to the page slug?  I suppose it doesn't matter too much now since we're moving to autogenerated ids, but I've seen that in other places and am curious why we're doing that?  Seems counter-intuitive.

packages/context/src/app/AppProvider.tsx
```diff
@@ -62,8 +78,57 @@ export type AppProviderProps = {
...
+                        : defaultPageId;
+                    defaultPage = await apiServices.get<Page>(
+                        `/webContent/apps/${app.id}/pages/${encodeURIComponent(encodeURIComponent(pageId))}`,
```
actong, line 122: Have you tested that this works?  Not sure if this will generate `https://...evokeplatform.com/api/webContent/...` or `https://...evokeplatform.com/webContent/...`.  BTW would be nice to have unit tests.

bradwray: Yeah, it works. It ends up with /api on it. We add that to the baseUrl before passing that to the ApiBaseUrl Provider.

packages/context/src/app/AppProvider.tsx
bradwray: That's how the default pages are saved. I'm not sure why that was decided. Why is it counter-intuitive?

packages/context/src/app/AppProvider.tsx
bradwray: Good point, I'll get rid of that typeof check.

packages/context/src/app/AppProvider.tsx
```diff
@@ -62,8 +84,48 @@ export type AppProviderProps = {
...
+
+                    currentObjectId = effectiveObject?.baseObject?.objectId ?? undefined;
+                }
+
+                let defaultPage: Page | undefined;
+                if (defaultPageId) {
+                    const pageId = defaultPageId.includes('/')
+                        ? defaultPageId.split('/').slice(2).join('/')
+                        : defaultPageId;
+                    defaultPage = await apiServices.get<Page>(
```
actong, line 116: I think we can catch 404 errors here.  If the page doesn't exist, probably shouldn't make the widget fail, just treat it as if there is no default page.

packages/context/src/app/AppProvider.tsx
actong: It's not part of the page id/slug.  I noticed it with `navigateTo`, I think it expects you to include the app prefix in the navigation target, but that made for an awkward api for widget authors.  Most of the time you're navigating within the same app, so a widget that needs to do some navigation first has to figure out what app it's currently in, get the slug for that app, and then prepend that to the page slug.

packages/context/src/app/AppProvider.tsx
```diff
@@ -113,9 +114,16 @@ function AppProvider(props: AppProviderProps) {
...
-                    defaultPage = await apiServices.get<Page>(
-                        `/webContent/apps/${app.id}/pages/${encodeURIComponent(encodeURIComponent(pageId))}`,
-                    );
+                    try {
+                        defaultPage = await apiServices.get<Page>(
+                            `/webContent/apps/${app.id}/pages/${encodeURIComponent(encodeURIComponent(pageId))}`,
+                        );
+                    } catch (error) {
+                        const err = error as AxiosError;
+                        if (err.status === 404) {
```
actong, line 123: `AxiosError.status` is a string, and I'm not too sure what gets populated there.  I think it's safer to check the status on `AxiosError.response`.

packages/context/src/app/AppProvider.tsx
```diff
@@ -50,8 +53,28 @@ export type NavigationItem = {
...
 
-const defaultApp: App = { id: '_evoke', name: 'Evoke Platform', type: 'public' };
-const AppContext = createContext<App>(defaultApp);
+const defaultApp: App = {
+    id: '_evoke',
+    name: 'Evoke Platform',
+    type: 'public',
+};
+
+type AppExtended = App & {
```
actong, line 62: Should probably export this type too.

PR #437, opened 2024-06-14
fix: update widget properties for sample widget

bradwray:
The widget properties of the sample widget were not aligned with how the builder reads them. This fixes that and provides a better learning experience for new widget developers..

### DIFF
--- a/packages/plugin/src/generators/app/templates/src/widgets/SampleWidget/WidgetProperties.json
+++ b/packages/plugin/src/generators/app/templates/src/widgets/SampleWidget/WidgetProperties.json
@@ -6,10 +6,10 @@
     "src": "src/widgets/SampleWidget",
     "properties": [
         {
-            "id": "message",
-            "name": "Message",
+            "name": "message",
             "type": "text",
-            "optional": true
+            "isOptional": true,
+            "inputLabel": "Message"
         }
     ]
 }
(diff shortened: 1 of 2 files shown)


packages/plugin/src/generators/app/templates/src/widgets/SampleWidget/WidgetProperties.json
```diff
@@ -6,10 +6,10 @@
     "src": "src/widgets/SampleWidget",
     "properties": [
         {
-            "id": "message",
-            "name": "Message",
+            "name": "message",
             "type": "text",
-            "optional": true
+            "isOptional": true,
+            "inputLabel": "Message"
```
actong, line 12: What is this used for, and how is it different from `displayName`?

bradwray: You're right. Should be displayName. I'll change it.

bradwray:
When running the generated widget in storybook, an error showing that the "@babel/plugin-proposal-class-properties" dependency was not found. Installing it allows storybook to work again.

actong:
These commits should probably be types as `fix`, not `chore`.

bradwray:
> These commits should probably be types as `fix`, not `chore`.

Is there any simpler way to change that other than the whole rebase reword way in vim?

actong:
Not sure what you mean by "reword way in vim", but you'd have to rebase the commits to change the message.

PR #446, opened 2024-06-20
feat: redirect to logout screen

mtruongsa:
When log out successfully, users will be redirected to the logout screen.
Users will be redirected to the last page before they logged out when they log in again.

### DIFF
--- a/packages/context/src/authentication/AuthenticationContextProvider.tsx
+++ b/packages/context/src/authentication/AuthenticationContextProvider.tsx
@@ -52,17 +52,17 @@ function AuthenticationContextProvider(props: AuthenticationContextProviderProps
         () =>
             account
                 ? {
-                    account: { id: account.localAccountId, name: account.name },
-                    logout: () => {
-                        msal.instance.logoutRedirect({
-                            account,
-                            postLogoutRedirectUri:
-                                `/logout?p=${encodeURIComponent(window.location.pathname + window.location.search)}`
-
-                        });
-                    },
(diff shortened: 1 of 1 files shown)


packages/context/src/authentication/AuthenticationContextProvider.tsx
```diff
@@ -54,7 +54,14 @@ function AuthenticationContextProvider(props: AuthenticationContextProviderProps
                 ? {
                       account: { id: account.localAccountId, name: account.name },
                       logout: () => {
-                          msal.instance.logoutRedirect({ account });
+                          msal.instance.logoutRedirect({
+                              account,
+                              postLogoutRedirectUri:
+                                  '/logout' +
+                                  (window.location.pathname !== '/'
```
actong, line 61: Doesn't look like `pathname` would ever be `/` (https://developer.mozilla.org/en-US/docs/Web/API/Location/pathname#:~:text=If%20there%20is%20no%20path%2C%20pathname%20will%20be%20empty), though even if it was, is there some problem with passing `?p=%2F` (p=/) that this check is trying to prevent?

mtruongsa: If i am at mtruong.cedar.mylicenseone.com, then window.location.pathname always returns '/'. It is not a problem. I was just trying to make it cleaner. However, I think we will need it because we will need to add query parameters if the last page users logout does have query parameter.

actong (approved):

PR #482, opened 2024-07-29
feat: add initialValue object type

smark-1:
Makes `initialValue` using an object type valid. 

Example
```json
{
    "name": "action",
    "type": "inputGroup",
    "displayName": "Action",
    "isOptional": true,
    "initialValue": {"label": "green", "favoriteColor": "yellow"}, // now valid
    "inputGroup": [
        {
            "type": "text",
            "name": "label",
            "displayName": "Text Color"
        },
        {
            "type": "text",
            "name": "label",
            "displayName": "Background Color"
        }
    ]
}
```

### DIFF
--- a/widgetschema.json
+++ b/widgetschema.json
@@ -187,7 +187,7 @@
                     "type": "boolean"
                 },
                 "initialValue": {
-                    "oneOf": [{ "type": "string" }, { "type": "number" }, { "type": "boolean" }],
+                    "oneOf": [{ "type": "string" }, { "type": "number" }, { "type": "boolean" }, { "type": "object" }],
                     "description": "Initial value of the property"
                 },
                 "tooltip": {


actong:
What does it mean for an `inputGroup` to have an initial value?  I'm not really following the posted example.

smark-1:
Sorry I took the example from the docs. It appears to be wrong. It should be the below. The key corresponds to the name of the inputGroup item and will set the default value of it to the value of the key. 
```JSON
 {
            "name": "action",
            "type": "inputGroup",
            "displayName": "Action",
            "isOptional": true,
            "initialValue": {"label": "green", "favoriteColor": "yellow"},  //now valid
            "inputGroup": [
                {
                    "type": "text",
                    "name": "label",
                    "displayName": "Text Color"
                },
                {
                    "type": "text",
                    "name": "favoriteColor",
                    "displayName": "Background Color"
                }
            ]
        }
```

actong (approved):

PR #508, opened 2024-09-10
fix: fully specify import for useApiServices

mk160107:
The import of `useApiServices` in `AppProvider` did not contain a valid file extension. This was causing an issue where `useApiServices` would fail to resolve from `../api` when running `pkgplugin` for some custom widget projects.

### DIFF
--- a/packages/context/src/app/AppProvider.tsx
+++ b/packages/context/src/app/AppProvider.tsx
@@ -2,7 +2,7 @@
 // This file is licensed under the MIT License.
 
 import { createContext, ReactNode, useCallback, useContext } from 'react';
-import { useApiServices } from '../api';
+import { useApiServices } from '../api/index.js';
 import { Obj } from '../objects';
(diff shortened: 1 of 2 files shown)


packages/context/src/app/AppProvider.tsx
```diff
@@ -2,7 +2,7 @@
 // This file is licensed under the MIT License.
 
 import { createContext, ReactNode, useCallback, useContext } from 'react';
-import { useApiServices } from '../api';
+import { useApiServices } from '../api/index.js';
 import { Obj } from '../objects';
```
actong, line 6: This also needs updating.  Probably need to go through the entire project, `module` packages can't import from a directory, you have to specify the file including the extension.

mk160107: Good catch. I went through every file in the context and double checked that the extensions were included.

actong (approved):

PR #528, opened 2024-10-03
feat: remove SignalRConnectionProvider implementation

actong:
This feature is no longer supported and was superseded by NotificationProvider.  Note that while this is not a breaking change to any code making use of this library (meaning the code won't fail), consumers of useSignalRConnection will no longer receive any notifications of changes.


packages/context/README.md
```diff
@@ -206,101 +206,12 @@ absolute URL.
...
-instanceChanges.unsubscribe('myObjectId', callback);
-```
-
-### Notification
+### Notifications
 
 -   [useNofitication](#usenotification)
     -   [documentChanges](#documentchangessubscribeobjectidinstanceid-data-documentchange)
     -   [instanceChanges](#instancechangessubscribeobjectid-instanceids-instancechange)
+-   [~~useSignalRConnection~~](#usesignalrconnection) (_Deprecated_)
```
rverma3, line 214: @actong We might not want to use the term "deprecated" for it -- it implies that it still functions as before but can become non-functional later so its use is not encouraged. From the description below it seems like it doesn't really work as intended any more, it just won't be breaking any code (which we anyways have to ensure).

I think we normally use "Obselete" for such cases. Please see this for reference: https://www.baeldung.com/cs/deprecated-vs-obsolete

packages/context/src/signalr/SignalRConnectionProvider.tsx
```diff
@@ -35,142 +33,17 @@ export const SignalRConnectionContext = createContext<SignalRConnectionContextTy
...
-                    ? {
-                          subscribe: (topicName, callback) => instancesSignalRConnection.on(topicName, callback),
-                          unsubscribe: (topicName, callback) =>
-                              callback
-                                  ? instancesSignalRConnection.off(topicName, callback)
-                                  : instancesSignalRConnection.off(topicName),
-                      }
-                    : undefined,
+                documentChanges: {
+                    subscribe: () => {},
```
rverma3, line 40: @actong Where is the function logic being moved to?

rverma3 (changes requested):

packages/context/src/signalr/SignalRConnectionProvider.tsx
actong: Not following, what do you mean by function logic?

rverma3: where are we defining the subscribe / unsubscribe methods?

packages/context/src/signalr/SignalRConnectionProvider.tsx
```diff
@@ -179,6 +52,8 @@ function SignalRConnectionProvider({ children }: { children: React.ReactNode })
 }
 
 export function useSignalRConnection() {
+    console.warn('Use of useSignalRConnection is deprecated. Use useNotification instead.');
```
rverma3, line 55: Please update here as well.

rverma3 (changes requested):

packages/context/src/signalr/SignalRConnectionProvider.tsx
actong: These are the subscribe/unsubscribe functions.  Or if you mean the type definition, that's defined above by `SignalRConnectionContextType`.

rverma3: oh nvm. I realise now this is just for backward compatibility.

rverma3 (approved):

packages/context/src/signalr/SignalRConnectionProvider.tsx
actong: I suppose another option would be to have `useSignalRConnection` actually use `useNotification` under the hood, in which case it would still be functional.

rverma3: that's a good idea -- not sure if I would hold this up for that though

actong: I went ahead and implemented it since this means we don't have to worry so much about warning users to change their widgets.  Note I changed the verbiage back to deprecated.

packages/context/src/signalr/SignalRConnectionProvider.tsx
```diff
@@ -33,16 +34,39 @@ export const SignalRConnectionContext = createContext<SignalRConnectionContextTy
...
 function SignalRConnectionProvider({ children }: { children: React.ReactNode }) {
+    const notifications = useNotification();
+
     return (
         <SignalRConnectionContext.Provider
             value={{
                 documentChanges: {
-                    subscribe: () => {},
-                    unsubscribe: () => {},
+                    subscribe: (topicName, callback) => {
```
Richa312, line 43: @actong what's a topicName?

rverma3, line 43: @actong is topicName essentially the objectID for the object this document is attached to?

rverma3: nvm, i see its the "objectid/instanceid" string... is this the signature we've had previously?

actong: For documents yes, for instances it's just the objectId.  See current documentation https://github.com/Evoke-Platform/evoke-sdk/blob/033bccad4d9da9aa55816c894e4137a73e95ea34/packages/context/README.md#signalr-connection

packages/context/README.md
```diff
@@ -306,8 +308,99 @@ instanceChanges.unsubscribe('myObjectId', callback);
...
+
+##### `documentChanges.unsubscribe('{objectId}/{instanceId}', (data: DocumentChange[]) => {})`
+
+> **Deprecated**
+> This has been deprecated in favor of [useNotification](#usenofitication).
+
+Unsubscribe to the specified object instance document changes.
+
+Callback function is optional.
+If callback function is defined, you must pass the exact same Function instance as was previously passed to `documentChanges.subscribe`.
```
rverma3, line 350: @actong not sure if this was the same before as well, but it seems to me this sentence needs to be highlighted more / marked important. What do you think?

actong: The PR diff makes it hard to see, but the content is the same as before (other than I added the **Deprecated** notices), I just moved `useSignalRConnection` below `useNotification` (didn't want the deprecated version to come first).

As for marking it more important, I don't have any strong feelings one way or the other.  I think it's pretty standard that when unsubscribing/removing a listener, you'd have to pass the actual listener you want to unsubscribe.

packages/context/src/tests/signalr/useSignalRConnection.unit.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+        // Simulate a notification.
+        callback?.(
+            { objectId: 'docObject', instanceId: 'testInstance', documentId: 'doc1', type: 'type1' },
+            { objectId: 'docObject', instanceId: 'testInstance', documentId: 'doc2', type: 'type2' },
+        );
+
+        sinon.verify();
+    });
+});
```
rverma3, line 172: @actong do we also want to test if `unsubscribe` methods are being called properly? Also I don't know how in-depth we want to go, but maybe also test for using the same instance of the callback function?

actong: Hmm, yeah, good point, right now it wouldn't work for instances since it's always a new function.

rverma3 (changes requested):
marking it as "changes requested" for the test cases

packages/context/src/tests/signalr/useSignalRConnection.unit.tsx
rverma3: I wonder if there's a way to still use useNotification with instances here without impact backwards compatibility.

rverma3 (approved):

packages/context/src/signalr/SignalRConnectionProvider.tsx
```diff
@@ -34,143 +33,57 @@ export const SignalRConnectionContext = createContext<SignalRConnectionContextTy
...
-        }
-
-        return () => {
-            instancesSignalRConnection?.stop();
-            instancesConnectionStopped = true;
-        };
-    }, [instancesSignalRConnection]);
+    const notifications = useNotification();
+
+    const [instanceCallbacks] = useState(
```
rverma3, line 42: @actong I approved it cuz this would work, but do we need useState here, or can we just put this in a constant?

actong: If it was just a local const, it would get reset every time the component re-renders.  Or you mean it'd be a const at the module level?

PR #529, opened 2024-10-30
feat: upgrade axios version in context library

mtruongsa:


packages/context/package.json
```diff
@@ -68,7 +68,7 @@
     },
     "dependencies": {
         "@microsoft/signalr": "^7.0.12",
-        "axios": "^1.6.7",
+        "axios": "1.7.3",
```
bradwray, line 71: I see this https://security.snyk.io/package/npm/axios

Why just upgrade to 1.7.3?

mtruongsa: I did the latest version then it causes the tests to fail with Invalid URL. 
There is a breaking change in 1.7.4+ versions. https://github.com/axios/axios/issues/6580

mk160107 (approved):

jhowdenSA (approved):

PR #530, opened 2024-11-21
feat: cdr-1632 add table view layout type

mtruongsa:


packages/context/src/objects/objects.ts
```diff
@@ -11,20 +11,40 @@ export type BaseObjReference = {
     discriminatorValue: unknown;
 };
 
-export type ViewLayout = {
-    table?: TableViewLayout;
-    dropdown?: DropdownViewLayout;
+export type DefaultViewLayout = {
```
actong, line 14: Renaming this is a breaking change.

PR #532, opened 2024-11-25
feat: add new version of layout types

mtruongsa:


packages/context/src/objects/objects.ts
```diff
@@ -11,6 +11,21 @@ export type BaseObjReference = {
     discriminatorValue: unknown;
 };
 
+export type ViewLayoutReference = {
+    id: string;
+    objectId: string;
+};
+
+export type ViewLayoutV2 = {
```
actong, line 19: How would these be used?

mtruongsa: In the configuration repo, we will be using TableViewLayoutV2 and DropdownViewLayoutV2 when calling api endpoints. ViewLayoutV2 is a base for these two types. Maybe we don't need to export it.

actong: Hmm, ok.  Was thinking these didn't need to be here at all (context library doesn't export types for everything, e.g. correspondence templates), but I guess `ViewLayoutReference` sort of references these.

I guess I'm ok with the `V2` suffix, though part of me feels it isn't quite appropriate since they're used differently.  Not sure if there's some more descriptive name like `ViewLayoutStandalone` or `ViewLayoutEntity`, though the name becomes a bit long.

The other thing is maybe `ViewLayoutReference` should match the new names, otherwise you might think it's a reference to `ViewLayout`.

mtruongsa: If Scarlet starts doing the configurations then I think it is good we have the types here for custom widgets to use.

actong (approved):

PR #533, opened 2024-11-26
feat: update action input type

mtruongsa:


packages/context/src/objects/objects.ts
```diff
@@ -62,6 +62,11 @@ export type Obj = {
 
 export type ObjWithRoot = Obj & { rootObjectId: string };
 
+export type EffectiveObject = Obj & {
```
actong, line 65: Why is this needed?  I don't think there's any api that returns this.

packages/context/src/objects/objects.ts
```diff
@@ -323,9 +332,9 @@ export type ActionInput = {
         when?: string;
         eq?: string | number | boolean;
     };
-    property?: {
-        id: string;
-    };
+    property?: InputParameter;
+    viewLayout?: ViewLayoutEntityReference;
+    documentMetadata?: DocumentMetadata;
```
actong, line 337: Is this the legacy `inputProperties`?  Are we still maintaining this?

actong:
What are these changes for?

packages/context/src/objects/objects.ts
mtruongsa: you are right. this is only used locally in Configuration repo to buildEffectiveObjects.

packages/context/src/objects/objects.ts
mtruongsa: Yes it is legacy `inputProperties`. 
I am using this type `ActionInput` instead of the one in Configuration repo. The one called FormIOComponent is in the Configuration repo is removed. 
There are too many duplicated types between this repo and configuration repo. So I am removing all those duplicated ones in Configuration repo to use the ones in this repo as a single source of truth.

mtruongsa: Also, we are using `ActionInput` type when we convert form to formio component in Widget repo.

mtruongsa:
> What are these changes for?

somehow my local always changes to this when I do git commit.

packages/context/src/objects/objects.ts
```diff
@@ -278,6 +275,13 @@ export type ActionInputType =
     | 'Time'
     | 'User';
 
+export type DocumentType = 'Portal' | 'Private' | 'Public';
+
+export type DocumentMetadata = {
```
actong, line 280: Document metadata is open-ended.  As far as the platform is concerned, this is just `Record<string, string>`.

actong:
> > What are these changes for?
> 
> somehow my local always changes to this when I do git commit for this repo.

I was asking about the PR as a whole.

packages/context/src/objects/objects.ts
```diff
@@ -278,6 +275,8 @@ export type ActionInputType =
     | 'Time'
     | 'User';
 
+export type DocumentType = 'Portal' | 'Private' | 'Public';
```
actong, line 278: This is not needed either.

packages/context/src/objects/objects.ts
```diff
@@ -323,9 +322,9 @@ export type ActionInput = {
         when?: string;
         eq?: string | number | boolean;
     };
-    property?: {
-        id: string;
-    };
+    property?: InputParameter;
+    viewLayout?: ViewLayoutEntityReference;
```
actong, line 326: Why is this being added to the legacy `inputProperties`?  When is this ever set?

mtruongsa: We have a function in Configuration and Widget repos to convert our form to form io components. The function returns a list of ActionInput components. To access `viewLayout` in those components, I need to add to `ActionInput` type.

actong (approved):

PR #535, opened 2024-12-18
feat: add custom params serializer to api services

mtruongsa:


actong:
Please add tests.  I suspect this breaks simple string parameters.

packages/context/src/tests/api/api.unit.ts
```diff
@@ -134,10 +138,35 @@ describe('ApiServices', () => {
 
         describe('(url, config) => Promise', () => {
```
actong, line 139: Please test setting a param to `undefined`.

actong:
Note if you're just updating unit tests, that's a `chore`, not a `feat`ure.

actong (approved):

PR #537, opened 2025-01-14
feat: filter out undefined params

mtruongsa:


packages/context/src/tests/api/api.unit.ts
```diff
@@ -150,6 +150,12 @@ describe('ApiServices', () => {
                 expect(data).to.eql('');
             });
 
+            it('sends undefined params value', async () => {
```
actong, line 153: It's actually **not** sending.

packages/context/src/api/paramsSerializer.ts
```diff
@@ -4,7 +4,9 @@ export function paramsSerializer(params: Record<string, unknown>, options?: Para
     const searchParams = new URLSearchParams();
 
     for (const [key, value] of Object.entries(params)) {
-        searchParams.append(key, typeof value !== 'string' ? JSON.stringify(value) : value);
+        if (key && value) {
```
actong, line 7: This will ignore any falsy value.

actong (changes requested):

packages/context/src/api/paramsSerializer.ts
```diff
@@ -1,11 +1,25 @@
...
 
     return searchParams.toString();
 }
+
+function validateParam(item: unknown) {
+    return (
+        !_.isNil(item) &&
+        !_.isUndefined(item) &&
+        !(_.isObject(item) && _.isEmpty(item)) &&
+        !(_.isString(item) && _.isEmpty(item))
```
actong, line 23: Why are empty objects and empty strings not valid?  Empty strings are definitely valid and a common occurrence.  The only thing that is clear to me that shouldn't be sent is `undefined`.  Anything else seems like a legitimate value.

actong:
Please use the appropriate commit type in the message.  Should use `fix` for bugfixes.

packages/context/src/api/paramsSerializer.ts
```diff
@@ -1,11 +1,24 @@
...
+        }
+
         searchParams.append(key, typeof value !== 'string' ? JSON.stringify(value) : value);
     }
 
     return searchParams.toString();
 }
+
+function validateParamKey(item?: string) {
+    return !_.isNil(item) && !_.isUndefined(item) && !(_.isString(item) && _.isEmpty(item));
```
actong, line 19: Keys are already strings.  Not sure how much value there is to validating the key.  The only thing that's arguably invalid is `''`, and even then not sure it's actually invalid, though I can see there's little reason to support an empty key so I think it's fine if we ignore it.

packages/context/src/api/paramsSerializer.ts
```diff
@@ -1,11 +1,24 @@
 import { ParamsSerializerOptions } from 'axios';
+import _ from 'lodash';
```
actong, line 2: Don't use lodash for browser.  lodash-es allows webpack to do tree shaking, though having said that, there's no reason to use lodash to check if something is undefined.

actong (approved):

PR #538, opened 2025-01-16
feat: add multiple widget property in json schema

tlachman:


widgetschema.json
```diff
@@ -228,8 +233,53 @@
                             "items": { "$ref": "#/definitions/dataType" }
                         }
                     }
+                },
+                "visibility": { "$ref": "#/definitions/WidgetVisibilityGroup" },
+                "fields": {
+                    "type": "array",
+                    "items": { "$ref": "#/definitions/widgetProperty" }
+                },
+                "itemName": {
```
actong, line 242: I'm assuming `fields` are the properties of the repeated nested objects inside a "multiple" property, but I'm not following what `itemName` is.

Also, do `fields` and `itemName` have any meaning outside of a "multiple" property?  If not, should these be defined in the "multiple" definition above in `oneOf`?

actong: I see.  So a multiple property might have a plural form like "Buttons", and then `itemName` would have the singular form "Button"?

One concern is that `itemName` might be seen as similar to `name`, i.e. a machine-readable id, but really it seems to be more like `displayName`, a label for presentation purposes only.  I also have trouble reconciling that `displayName` is optional but `itemName` is required.  I would think they're both optional.

widgetschema.json
```diff
@@ -115,6 +115,10 @@
                         }
                     },
                     "description": "Mapping of the API result to the choices"
+                },
+                "mappedValues": {
+                    "type": "object",
+                    "description": "Values from other widget properties that will replace parts of the url. The key specifies the url segment to replace, and the value indicates the corresponding property name."
```
actong, line 121: Couple questions:

1. Will this be able to access properties inside input groups, e.g. `someInputGroup.somePropInTheInputGroup`?
2. If the `choices` property is inside a `multiple`, does the value reference here relative to the current item inside the `multiple` or is it relative to the root widget properties?  Or preferably, is there a way to get to either one depending on what's desired?

widgetschema.json
```diff
@@ -162,6 +195,7 @@
                         "choices",
                         "inputGroup",
                         "date",
+                        "multiple",
```
actong, line 198: I see we already have a `isMultiple` flag.  How does that relate to this new `multiple` type?  Or how is `multiple` different from `inputGroup` with `isMultiple = true`?

widgetschema.json
```diff
@@ -130,21 +134,50 @@
...
-                { "properties": { "type": { "const": "boolean" } } },
                 {
                     "properties": { "type": { "const": "choices" } },
-                    "oneOf": [{ "required": ["name", "type", "choices"] }, { "required": ["name", "type", "api"] }]
+                    "oneOf": [{ "required": ["choices"] }, { "required": ["api"] }],
+                    "exclude": ["fields", "itemDisplayName", "inputGroup"]
+                },
+                {
+                    "properties": { "type": { "const": "multiple" } },
+                    "required": ["fields", "itemDisplayName"],
```
actong, line 163: I guess I still don't think `itemDisplayName` should be required.  Does something not work if there is no `itemDisplayName`?

widgetschema.json
```diff
@@ -130,21 +134,50 @@
             "oneOf": [
                 {
                     "properties": { "type": { "const": "inputGroup" } },
-                    "required": ["name", "type", "inputGroup"]
+                    "required": ["name", "type", "inputGroup"],
+                    "exclude": ["fields", "itemDisplayName"]
```
actong, line 138: Is `exclude` part of json schema?

In general, this seems hard to maintain since any new property type would need to include this `exclude`.  I was thinking `fields` and `itemDisplayName` would be defined inside of `multiple` and not be in `widgetProperty`, and therefore would not need to be excluded from anything, though if that doesn't work either, I'm fine with the original option of just leaving them in `widgetProperty` so any property can specify them, even though in practice it wouldn't actually matter other than for `multiple`.

widgetschema.json
```diff
@@ -228,8 +262,54 @@
...
+                    "items": { "$ref": "#/definitions/widgetVisibilityCondition" }
+                }
+            },
+            "required": ["operator", "conditions"]
+        },
+        "widgetVisibilityCondition": {
+            "type": "object",
+            "description": "Condition for the visibility of a widget property",
+            "properties": {
+                "field": { "type": "string" },
```
actong, line 297: This is a reference to a widget property, similar to the values in `mappedValues`?

widgetschema.json
actong: I guess I do like that idea.  I'll definitely say that having a `multiple` type and a `isMultiple` flag that are completely unrelated seems confusing.

PR #552, opened 2025-01-24
chore: release v1.0.0 packages

mtruongsa:
context v1.0.0

payment v1.0.0

plugin-scripts v1.0.0

plugin v1.0.0

sdk v1.0.0

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,44 @@
+name: Release
+
+on:
+    workflow_dispatch:
+
+concurrency:
+    group: branch-live
+
+jobs:
+    publish:
+        runs-on: ubuntu-latest
+        if: github.ref == 'refs/heads/live'
+        steps:
+            - name: Generate Token
+              id: generate_token
+              uses: tibdex/github-app-token@v1
+              with:
+                  app_id: ${{ vars.TOKEN_APP_ID }}
+                  private_key: ${{ secrets.TOKEN_APP_PRIVATE_KEY }}
+            - uses: actions/checkout@v3
+              with:
+                  fetch-depth: 5
+                  token: ${{ steps.generate_token.outputs.token }}
+            - uses: actions/setup-node@v3
+              with:
+                  node-version: 18
+            - name: Setup
+              run: yarn install
+            - name: Build
+              run: yarn build
+            - name: Publish
+              run: |
+                  git config user.name "GitHub Actions Bot"
+                  git config user.email "<>"
+                  mv npmrc-ci .npmrc
+                  npx turbo run release --concurrency 1 --
+                  npx turbo run deploy --
+              env:
+                  NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+            - name: Push Changes
+              run: |
+                  git fetch origin
+                  git merge -m "Merge branch 'live' [skip ci]" origin/live
+                  git push origin HEAD:live
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,12 +4,12 @@ on:
     workflow_dispatch:
 
 concurrency:
-    group: branch-live
+    group: branch-release
 
 jobs:
     publish:
         runs-on: ubuntu-latest
-        if: github.ref == 'refs/heads/live'
+        if: github.ref == 'refs/heads/release'
         steps:
             - name: Generate Token
               id: generate_token
@@ -40,5 +40,5 @@ jobs:
             - name: Push Changes
               run: |
                   git fetch origin
-                  git merge -m "Merge branch 'live' [skip ci]" origin/live
-                  git push origin HEAD:live
+                  git merge -m "Merge branch 'release' [skip ci]" origin/release
+                  git push origin HEAD:release
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,6 @@ jobs:
                   private_key: ${{ secrets.TOKEN_APP_PRIVATE_KEY }}
             - uses: actions/checkout@v3
               with:
-                  fetch-depth: 5
                   token: ${{ steps.generate_token.outputs.token }}
             - uses: actions/setup-node@v3
               with:
@@ -41,4 +40,4 @@ jobs:
               run: |
                   git fetch origin
                   git merge -m "Merge branch 'release' [skip ci]" origin/release
-                  git push origin HEAD:release
+                  git push --follow-tags
--- a/.github/workflows/testing.continuous-integration.yml
+++ b/.github/workflows/testing.continuous-integration.yml
@@ -3,11 +3,11 @@ name: Testing Continuous Integration
 on:
     push:
         branches:
-            - live
+            - release
     workflow_dispatch:
 
 concurrency:
-    group: branch-live
+    group: branch-release
 
 jobs:
     publish:
@@ -46,5 +46,5 @@ jobs:
             - name: Push Changes
               run: |
                   git fetch origin
-                  git merge -m "Merge branch 'live' [skip ci]" origin/live
-                  git push origin HEAD:live
+                  git merge -m "Merge branch 'release' [skip ci]" origin/release
+                  git push origin HEAD:release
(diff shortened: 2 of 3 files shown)


mtruongsa:
There will be another PR to update the dependencies in sdk package.

actong:
There's nothing in this PR.  I would have expected something like a Github Actions workflow for doing a release.

mtruongsa:
> There's nothing in this PR. I would have expected something like a Github Actions workflow for doing a release.

**chore!** will force it to release a major version. 
https://www.conventionalcommits.org/en/v1.0.0/

actong:
What are you proposing the release process would look like?

.github/workflows/testing.continuous-integration.yml
```diff
@@ -0,0 +1,50 @@
+name: Continuous Integration
+
+on:
+    push:
+        branches:
+            - release/live
```
actong, line 6: So we're calling this a "release" branch?  Not saying it doesn't work, though want to make sure this won't be confusing to anyone.  I do think we don't need `/live`, not sure what that means.

.github/workflows/continuous-integration.yml
actong, line 1: Regarding the build process (https://github.com/Evoke-Platform/evoke-sdk/pull/552/files#diff-c471496bcb3ed47397b73d2e3e3aa41a8a679c86accaed9e05676fd9d5987434R28), this is for determining what changed when a PR is merged.  It's not going to work for a release build.  It probably doesn't work when merging from main to release branches either.

mtruongsa: How is it different if we merge something into main branch versus main/another branch merges into release branch? We still need to determine what changed so we can release new version of those changed packages, right? If we do not filter by baseline then we will update all the packages.

actong: When you do a release, you'll need to build anything that's changed since the last release, not just what's changed in the last commit.

packages/sdk/package.json
```diff
@@ -27,7 +27,7 @@
     "license": "MIT",
     "dependencies": {
         "@evoke-platform/context": "latest",
-        "@evoke-platform/payment": "^1.0.0-dev.0",
+        "@evoke-platform/payment": "latest",
```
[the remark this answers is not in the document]
mtruongsa: We need to release these packages first before we can update @evoke-platform/sdk dependencies.

.github/workflows/testing.continuous-integration.yml
```diff
@@ -0,0 +1,46 @@
...
+            - name: Setup
+              run: yarn install
+            - name: Build
+              run: yarn build
+            - name: Publish
+              run: |
+                  git config user.name "GitHub Actions Bot"
+                  git config user.email "<>"
+                  mv npmrc-ci .npmrc
+                  npx turbo run release --concurrency 1 -- --prerelease testing --skip.tag --skip.changelog
```
actong, line 38: So this builds and bumps the version on everything regardless of whether anything changed?

mtruongsa: Yes it will bump everything. should it be the same as `continuous-integration.yml` file?

actong: Ideally it's only building things that have changed since the last build, i.e. not generating a new version for something that didn't change and is exactly the same as the previous version.

Also, have you experimented locally on what this will do with future commits?  I tried simulating new changes and doing releases, but it keeps bumping the version to 2.0.0, similar to the behavior described in https://github.com/absolute-version/commit-and-tag-version/issues/210.

mtruongsa: It does not bump version to 2.0.0 for me after I do `fix and feat` commits and release.

actong: I didn't have a tag.  Once I had the 1.0.0 tag, it seemed to start bump minor/patch version numbers, though yeah, it doesn't understand monorepos so any change anywhere affects all packages.

mtruongsa:
@actong Based on the discussion yesterday, do you think I should also remove the filter in this main branch workflow? https://github.com/Evoke-Platform/evoke-sdk/pull/552/files#diff-85228f08b0080a2c556a4a1b84bdac72d0f82c29ac49645efa383033757ca168R36

.github/workflows/release.yml
```diff
@@ -0,0 +1,44 @@
...
+                  mv npmrc-ci .npmrc
+                  npx turbo run release --concurrency 1 --
+                  npx turbo run deploy --
+              env:
+                  NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+            - name: Push Changes
+              run: |
+                  git fetch origin
+                  git merge -m "Merge branch 'live' [skip ci]" origin/live
+                  git push origin HEAD:live
```
mtruongsa, line 44: @actong should I do a git checkout `main` and merge `live` to `main` so we don't have to do it manually?

.github/workflows/testing.continuous-integration.yml
```diff
@@ -0,0 +1,50 @@
+name: Testing Continuous Integration
+
+on:
+    push:
+        branches:
+            - live
```
actong, line 6: I thought this was going to be `release`?  Still not sure what "live" means.

.github/workflows/release.yml
actong: What is the plan for updating the versions in `main`?  If the release bumps for example the minor version, then all the versions in `main` will be wrong/stale.

.github/workflows/testing.continuous-integration.yml
mtruongsa: I tried to create `release` branch manually from a repo, it does not allow it.

.github/workflows/release.yml
mtruongsa: aren't the versions in `main` the same as `live` after the release? We will have to merge `live` back to `main`.

actong: `main` has (potentially) changes not in the release, in which case they would not be the same version.

actong: What will generate a new version?  I assume in this instance, merging into `main` will result in a conflict which has to be manually resolved.

.github/workflows/release.yml
```diff
@@ -0,0 +1,44 @@
...
+        steps:
+            - name: Generate Token
+              id: generate_token
+              uses: tibdex/github-app-token@v1
+              with:
+                  app_id: ${{ vars.TOKEN_APP_ID }}
+                  private_key: ${{ secrets.TOKEN_APP_PRIVATE_KEY }}
+            - uses: actions/checkout@v3
+              with:
+                  fetch-depth: 5
```
actong, line 22: Don't need to fetch anything extra.  This was being used to set baseline.

.github/workflows/release.yml
```diff
@@ -0,0 +1,44 @@
...
+                  mv npmrc-ci .npmrc
+                  npx turbo run release --concurrency 1 --
+                  npx turbo run deploy --
+              env:
+                  NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+            - name: Push Changes
+              run: |
+                  git fetch origin
+                  git merge -m "Merge branch 'release' [skip ci]" origin/release
+                  git push origin HEAD:release
```
actong, line 44: Don't think this will push the tags.

packages/sdk/package.json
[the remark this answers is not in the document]
actong: What's the plan for dealing with this issue?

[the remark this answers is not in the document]
mtruongsa: I plan to have another PR just to update to the stable version.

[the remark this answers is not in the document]
actong: Meaning we're going to release 1.0.0 for sdk, then immediately release a 1.0.1 with just this change?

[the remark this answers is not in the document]
mtruongsa: Yes.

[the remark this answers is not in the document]
actong: I don't know.  Maybe this works when we don't have to update these dependencies, but at least for the initial 1.0.0, I think we'll need to do the release manually (without these workflows), I don't want to release a 1.0.1 just to update these versions.

PR #561, opened 2025-02-05
feat: add choicesDisplay and sortBy options for singleSelect styling

Meisosk:
Needed for making single select properties into radio buttons and sorting them according to their configuration/alphabetically

### DIFF
--- a/packages/context/src/objects/objects.ts
+++ b/packages/context/src/objects/objects.ts
@@ -211,7 +211,7 @@ export type DisplayConfiguration = {
     visibility?: VisibilityConfiguration | string;
     viewLayout?: ViewLayoutEntityReference;
     choicesDisplay?: {
-        type: 'dropdown' | 'radio';
+        type: 'dropdown' | 'radioButton';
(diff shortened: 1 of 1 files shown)


mtruongsa:
should `sortBy` values be `'none' | 'asc' | 'desc'`? I think we use these values for autocomplete.

packages/context/src/objects/objects.ts
```diff
@@ -210,6 +210,8 @@ export type DisplayConfiguration = {
     relatedObjectDisplay?: 'dropdown' | 'dialogBox';
     visibility?: VisibilityConfiguration | string;
     viewLayout?: ViewLayoutEntityReference;
+    choicesDisplay?: 'dropdown' | 'radio';
```
mtruongsa, line 213: maby turn this into an object that has `type: 'dropdown' | 'radio'; sort: 'none' | 'asc' | 'desc'`

mtruongsa (approved):

mtruongsa (approved):

PR #564, opened 2025-02-13
feat: add radioButton option to ActionInput displayOption

Meisosk:
This will stop the Configuration pipeline from failing

### DIFF
--- a/packages/context/src/objects/objects.ts
+++ b/packages/context/src/objects/objects.ts
@@ -306,7 +306,7 @@ export type ActionInput = {
     inputMaskPlaceholderChar?: string;
     tableView?: boolean;
     mode?: 'default' | 'existingOnly';
-    displayOption?: 'dropdown' | 'dialogBox';
+    displayOption?: 'dropdown' | 'dialogBox' | 'radioButton';
     rows?: number;
     showCharCount?: boolean;
     readOnly?: boolean;


mtruongsa:
We are showing collection and related object instances in a radio button now?

Meisosk:
> We are showing collection and related object instances in a radio button now?

No this is just to be able to set the display option to radioButton for single selects, but the way that value is saved in the component is as a displayOption same as a related object

actong:
It can certainly make sense for related instances (though whether we implement it is a different question).  Do collections make use of `displayOption`?

Meisosk:
I don't think collections use displayOption, it's only being applies to object property types as far as I can tell

packages/context/src/objects/objects.ts
```diff
@@ -306,7 +306,7 @@ export type ActionInput = {
     inputMaskPlaceholderChar?: string;
     tableView?: boolean;
     mode?: 'default' | 'existingOnly';
-    displayOption?: 'dropdown' | 'dialogBox';
+    displayOption?: 'dropdown' | 'dialogBox' | 'radioButton';
```
actong, line 309: Note this is the legacy form configuration.  The new form config is in `Form`.

mtruongsa:
> > We are showing collection and related object instances in a radio button now?
> 
> No this is just to be able to set the display option to radioButton for single selects, but the way that value is saved in the component is as a displayOption same as a related object

You mean when it converts from `form` to `formIO` to render, `choicesDisplay` is stored to `displayOption`?

Meisosk:
> > > We are showing collection and related object instances in a radio button now?
> > 
> > 
> > No this is just to be able to set the display option to radioButton for single selects, but the way that value is saved in the component is as a displayOption same as a related object
> 
> You mean when it converts from `form` to `formIO` to render, `choicesDisplay` is stored to `displayOption`?

Yes the choicesDisplay.type is stored there, and the choicesDisplay.sortBy is stored in the sortBy

packages/context/src/objects/objects.ts
Meisosk: Without this change the convertFormToComponents function in the ui-copmponent-lib form utils.ts  has an error. Is this function not used for current forms?

actong: I'm not saying this shouldn't change.  I'm saying the PR does not update the new form config.

mtruongsa: In his last PR, he added `choicesDisplay` object to `DisplayConfiguration` type.
I am not seeing it included in this PR but it does exist in the main branch.

actong: Ah, meaning this was already done in the new configuration, and this PR is just to update the legacy FormIO config?

mtruongsa: Yes

mtruongsa (approved):

PR #576, opened 2025-02-26
chore: add custom release workflow

mtruongsa:


mtruongsa:
After we merge this PR, we will invoke the workflow manually to publish `context`, `payment`, `plugin` and `plugin-scripts`. Then we have a new PR to update `sdk` dependencies and invoke the workflow manually again to publish `sdk`. 

The other PR can be merged with skip ci so it does not publish new version.

.github/workflows/custom-release.yml
```diff
@@ -0,0 +1,47 @@
...
+                  app_id: ${{ vars.TOKEN_APP_ID }}
+                  private_key: ${{ secrets.TOKEN_APP_PRIVATE_KEY }}
+            - uses: actions/checkout@v3
+              with:
+                  token: ${{ steps.generate_token.outputs.token }}
+            - uses: actions/setup-node@v3
+              with:
+                  node-version: 18
+            - name: Setup
+              run: yarn workspace @evoke-platform/${{ github.event.inputs.package_name }} install
```
actong, line 31: Use turbo so it can take care of building any dependency project.

.github/workflows/custom-release.yml
```diff
@@ -0,0 +1,47 @@
+name: Custom Release
+
+on:
+    workflow_dispatch:
+        inputs:
+            package_name:
```
actong, line 6: Could make this a `choice` type.

.github/workflows/custom-release.yml
```diff
@@ -0,0 +1,47 @@
...
+            - name: Setup
+              run: yarn workspace @evoke-platform/${{ github.event.inputs.package_name }} install
+            - name: Build
+              run: yarn workspace @evoke-platform/${{ github.event.inputs.package_name }} build
+            - name: Publish
+              run: |
+                  git config user.name "GitHub Actions Bot"
+                  git config user.email "<>"
+                  mv npmrc-ci .npmrc
+                  npx turbo run release --force --filter=@evoke-platform/${{ github.event.inputs.package_name }} --concurrency 1 --
```
actong, line 39: Why `--force`?  This will make it rebuild again.

actong (commented):
So this supersedes the other PR?  I.e. we run this workflow whenever we want to do a release?

mtruongsa:
> So this supersedes the other PR? I.e. we run this workflow whenever we want to do a release?

No. I think it is a one time thing.

.github/workflows/custom-release.yml
```diff
@@ -28,16 +35,16 @@ jobs:
               with:
                   node-version: 18
             - name: Setup
-              run: yarn workspace @evoke-platform/${{ github.event.inputs.package_name }} install
+              run: npx turbo run install --filter=@evoke-platform/${{ github.event.inputs.package_name }}
```
actong, line 38: `install` should be done at the root, otherwise it's not using the lock file.

actong:
> > So this supersedes the other PR? I.e. we run this workflow whenever we want to do a release?
> 
> No. I think it is a one time thing. Once we publish these packages, we can remove this workflow.

It could be the release workflow, I think.  If we want the other one that releases everything, though, then we don't need this one.  If it's one time, it doesn't need a github workflow.

PR #586, opened 2025-03-05
Resolve duplicate export

actong:
The new `Form` ui component conflicts with the pre-existing `Form` definition in object actions.  The [skip ci] prevented this from being caught in the previous PR.  For backwards compatibility, the pre-existing `Form` must stay, which means the ui component needs to be renamed.  This PR uses `RenderForm`, though I'm open to other suggestions.

### DIFF
--- a/packages/sdk/src/index.ts
+++ b/packages/sdk/src/index.ts
@@ -2,7 +2,7 @@
 // This file is licensed under the MIT License.
 
 export * from '@evoke-platform/context';
-export { Container as WidgetContainer } from '@evoke-platform/context';
+export { Form, Container as WidgetContainer } from '@evoke-platform/context';
 export * from '@evoke-platform/payment';
 export * from '@evoke-platform/ui-components';
 export { Container } from '@evoke-platform/ui-components';


mtruongsa (dismissed):

mtruongsa:
> The new `Form` ui component conflicts with the pre-existing `Form` definition in object actions. The [skip ci] prevented this from being caught in the previous PR. For backwards compatibility, the pre-existing `Form` must stay, which means the ui component needs to be renamed. This PR uses `RenderForm`, though I'm open to other suggestions.

Maybe name it `FormRenderer`?

mtruongsa:
> > The new `Form` ui component conflicts with the pre-existing `Form` definition in object actions. The [skip ci] prevented this from being caught in the previous PR. For backwards compatibility, the pre-existing `Form` must stay, which means the ui component needs to be renamed. This PR uses `RenderForm`, though I'm open to other suggestions.
> 
> Maybe name it `FormRenderer`?

Actually, I think RenderForm is ok. We will deprecate `Form` in the future. I think the new form will be named `FormRenderer`.

mtruongsa (approved):

PR #604, opened 2025-03-19
feat: add criteria property + support default value for criterias

tlachman:


actong:
I believe this should go to the `release` branch?

PR #625, opened 2025-04-01
chore: add authentication component

mtruongsa:
Refs: CDR-2171

### DIFF
--- a/packages/context/package.json
+++ b/packages/context/package.json
@@ -69,6 +69,7 @@
     "dependencies": {
         "@microsoft/signalr": "^7.0.12",
         "axios": "^1.7.9",
+        "react-idle-timer": "^5.7.2",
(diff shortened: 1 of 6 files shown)


actong:
Why is this being added to context?

actong:
Also, this should not be a `chore`

mtruongsa:
@actong I added here so we can use the same code in every repos instead of passing the components down as I did in previous PRs.

mtruongsa:
@actong if evoke context library is a singleton and shared across the platform, shouldn't Authenticate component be in this context library? Each remote app shouldn't have a different version of Authenticate component.

actong:
SDK is to help the development of widgets and payment gateways.  It's not a dumping ground for platform internals.

PR #626, opened 2025-04-04
feat: further define form types

mk160107:
This further defines the form types so that they align with the api specification.

### DIFF
--- a/packages/context/src/objects/objects.ts
+++ b/packages/context/src/objects/objects.ts
@@ -186,14 +186,16 @@ export type SelectOption = {
     value: string;
 };
 
+export type VisibilityCondition = {
+    property: string;
+    operator: 'eq' | 'neq';
+    value: string | number | boolean;
+    isInstanceProperty?: boolean;
+};
+
 export type VisibilityConfiguration = {
     operator?: 'any' | 'all';
-    conditions?: {
-        property: string;
-        operator: 'eq' | 'neq';
-        value: string | number | boolean;
-        isInstanceProperty?: boolean;
-    }[];
+    conditions?: VisibilityCondition[];
 };
 
 export type RelatedObjectDefaultValue = {
@@ -203,6 +205,8 @@ export type RelatedObjectDefaultValue = {
 };
 
 export type CriteriaDefaultValue = Record<string, unknown>;
+
+export type JsonLogic = Record<string, unknown> | boolean | number | string | null;
 
 export type DisplayConfiguration = {
     label?: string;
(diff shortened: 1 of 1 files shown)


packages/context/src/objects/objects.ts
```diff
@@ -248,19 +252,22 @@ export type Column = {
 
 export type Columns = {
     type: 'columns';
+    label?: string;
```
actong, line 255: What would this represent for `Columns`?

packages/context/src/objects/objects.ts
```diff
@@ -248,19 +252,22 @@ export type Column = {
...
     columns: Column[];
-    visibility?: VisibilityConfiguration | string;
+    visibility?: VisibilityConfiguration | JsonLogic;
 };
 
 export type Section = {
+    key?: string;
```
actong, line 261: How is `key` used?

packages/context/src/objects/objects.ts
```diff
@@ -204,10 +206,12 @@ export type RelatedObjectDefaultValue = {
 
 export type CriteriaDefaultValue = Record<string, unknown>;
 
+export type JsonLogic = Record<string, unknown> | boolean | number | string | null;
+
 export type DisplayConfiguration = {
     label?: string;
     placeholder?: string;
-    required?: boolean;
+    required?: boolean | string;
```
mtruongsa, line 214: should this have `boolean` type only?

packages/context/src/objects/objects.ts
mk160107: I'm not sure. It seems like we just use a random id in our form, although it is being used in the new form renderer.

packages/context/src/objects/objects.ts
mk160107: It doesn't represent anything at the moment, but it is configurable. 
image

packages/context/src/objects/objects.ts
mk160107: Yes, I'll fix this.

packages/context/src/objects/objects.ts
mk160107: Since this is not implemented in the form widget, should it be left out here?

packages/context/src/objects/objects.ts
```diff
@@ -204,10 +206,12 @@ export type RelatedObjectDefaultValue = {
 
 export type CriteriaDefaultValue = Record<string, unknown>;
 
+export type JsonLogic = Record<string, unknown> | boolean | number | string | null;
```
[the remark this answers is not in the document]
mk160107: Strangely enough, `boolean` and `number` are valid types for JSONLogic. You can see all the accepted values in the `RulesLogic` type here. https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/json-logic-js/index.d.ts. `string` is also valid, and we use it when the user saves invalid JSON.

packages/context/src/objects/objects.ts
actong: If there was something meaningful about it, I'd be ok adding it, but I just don't really see what this could mean.  To me a column layout is just about presentation, I don't think it needs a label.

mk160107: Makes sense. I've removed this property.

actong (approved):

PR #631, opened 2025-04-11
feat: update ActionInput type to include additional properties

ShermanE1:
The ActionInput type mimics the FormIO component model. This PR adds the properties that were added to the FormIO custom components, but are missing from the ActionInput model.

Refs: CDR-2140

### DIFF
--- a/packages/context/src/objects/objects.ts
+++ b/packages/context/src/objects/objects.ts
@@ -361,6 +361,17 @@ export type ActionInput = {
         maxDocuments?: number;
         customMessage?: string;
     };
+    isAddressLine1?: boolean;
+    addressPropertyId?: string;
+    defaultPages?: Record<string, string>;
+    instance?: ObjectInstance;
+    properties?: Property[];
+    canUpdateProperty?: boolean;
+    allCriteriaInputs?: string[];
+    apiServices?: ApiServices;
+    middleObject: ObjWithRoot;
+    initialMiddleObjectInstances: ObjectInstance[];
+    getMiddleObjectInstances: () => Promise<ObjectInstance[]>;
     /**
      * An array of sub-components to be rendered inside sections.
      */


packages/context/src/objects/objects.ts
```diff
@@ -361,6 +361,17 @@ export type ActionInput = {
         maxDocuments?: number;
         customMessage?: string;
     };
+    isAddressLine1?: boolean;
+    addressPropertyId?: string;
+    defaultPages?: Record<string, string>;
```
actong, line 366: We keep a copy of this on each form field?  Doesn't seem like the right place to store it.

actong (commented):
I pretty much had a question on every one of these new properties, but I have a feeling these are internal implementation details, is that correct?  If so I don't think they belong in the sdk, they should be kept internal.

packages/context/src/objects/objects.ts
mtruongsa: We don't store these in the database. They were added to form fields on fly to use.

packages/context/src/objects/objects.ts
```diff
@@ -361,6 +361,17 @@ export type ActionInput = {
         maxDocuments?: number;
         customMessage?: string;
     };
+    isAddressLine1?: boolean;
+    addressPropertyId?: string;
```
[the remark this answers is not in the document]
mtruongsa: `property.id` is something like `address.line1`, `address.line2`, etc... while `addressPropertyId` is just `address`

packages/context/src/objects/objects.ts
```diff
@@ -361,6 +361,17 @@ export type ActionInput = {
...
+    defaultPages?: Record<string, string>;
+    instance?: ObjectInstance;
+    properties?: Property[];
+    canUpdateProperty?: boolean;
+    allCriteriaInputs?: string[];
+    apiServices?: ApiServices;
+    middleObject: ObjWithRoot;
+    initialMiddleObjectInstances: ObjectInstance[];
+    getMiddleObjectInstances: () => Promise<ObjectInstance[]>;
```
mtruongsa, line 374: @ShermanE1 I think we should have a type like `type ExtendedActionInput = ActionInput & {...}` in the Form component instead of doing it here.

PR #633, opened 2025-05-01
feat: add boolean property to action input type

mtruongsa:
Refs: CDR-2115


yarn.lock
```diff
@@ -600,15 +600,6 @@
   resolved "https://registry.npmjs.org/@eslint/js/-/js-8.56.0.tgz#ef20350fec605a7f7035a01764731b2de0f3782b"
   integrity sha512-gMsVel9D7f2HLkBma9VbtzZRehRogVRfbr++f06nL2vnCGCNlzOD+/MUov/F4p8myyAHspEhVobgjpX64q5m6A==
 
-"@evoke-platform/context@^1.0.0":
```
mtruongsa, line 603: I did `yarn install` and this was removed. I don't know if this will be a problem for `sdk` package.

actong: I guess it wasn't updated on some prior PR.

Also, I see the released version of ui-components still depends on a prerelease version of context.

actong (approved):

yarn.lock
mtruongsa: I think because ui-components did not update the peer dependency to use the stable version of context.

PR #638, opened 2025-05-13
feat: add 'nonStrictEnum' to ActionInput model

ShermanE1:
The 'ActionInput' model replicates the FormIO input field model. Add 'nonStrictEnum' as a property there, so it can be used in the configuration of this property when the field is added to the form.

Refs: CDR-2238


actong:
Is this going to be supported, i.e. will `nonStrictEnum` work for legacy forms?  Otherwise if this was just because some code happens to be using this type in some internal implementation, this change should be localized to that piece of code.

PR #652, opened 2025-06-06
feat: add strictlyTrue flag for boolean property

mtruongsa:
Refs: CDR-2323

### DIFF
--- a/packages/context/src/objects/objects.ts
+++ b/packages/context/src/objects/objects.ts
@@ -136,7 +136,7 @@ export type Property = {
     id: string;
     name: string;
     type: PropertyType;
-    enum?: string[];
+    enum?: string[] | boolean[];
     nonStrictEnum?: boolean;
     objectId?: string;
     relatedPropertyId?: string;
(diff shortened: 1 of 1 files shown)


packages/context/src/objects/objects.ts
```diff
@@ -136,7 +136,7 @@ export type Property = {
     id: string;
     name: string;
     type: PropertyType;
-    enum?: string[];
+    enum?: string[] | boolean[];
```
aohsie, line 139: Will this not break any code that upgrades the SDK? Or is this just for scaffolding new projects?

mtruongsa: We will have to handle the types in the existing code where it is being used.

mtruongsa: Each custom widget can have its own evoke context version. If they update their version, they probably need to update the relevant code.

mtruongsa: The alternative is to have the local code extend this type. But for any custom widget that uses this type, they may not know that enum can be a ```boolean[]``` which can lead to type mismatch. Do you think it make sense to change this to ```any```?

aohsie (approved):

PR #658, opened 2025-06-23
feat(schema): disallow tooltip property for columnSelector widget property

bradwray:


widgetschema.json
```diff
@@ -162,7 +162,12 @@
                     "properties": { "type": { "const": "documentUpload" } }
                 },
                 {
-                    "properties": { "type": { "const": "columnSelector" } }
+                    "properties": { "type": { "const": "columnSelector" } },
+                    "not": {
```
actong, line 166: I tried this out at https://www.jsonschemavalidator.net/ and it didn't seem to work.  Using `tooltip: false` without the `not` seemed to work.

actong (approved):

PR #661, opened 2025-06-27
CDR-2308 unify MSAL and OIDC authentication context providers

aohsie:
- Add OIDC (react-oidc-context) support alongside MSAL, with auto-detection in AuthenticationContextProvider

### DIFF
--- a/packages/context/src/authentication/AuthenticationContextProvider.tsx
+++ b/packages/context/src/authentication/AuthenticationContextProvider.tsx
@@ -22,13 +22,13 @@ const Context = createContext<AuthenticationContext | undefined>(undefined);
 
 Context.displayName = 'AuthenticationContext';
 
-export type AuthenticationRequest = Pick<RedirectRequest, 'scopes' | 'extraQueryParameters' | 'state'>;
-
 export type AuthenticationContextProviderProps = {
     msal?: IMsalContext;
     authRequest: AuthenticationRequest;
     children?: ReactNode;
 };
+
+export type AuthenticationRequest = Pick<RedirectRequest, 'scopes' | 'extraQueryParameters' | 'state'>;
 
 function AuthenticationContextProvider(props: AuthenticationContextProviderProps) {
     // Auto-detect provider type based on presence of msal prop
@@ -94,7 +94,7 @@ function MsalProvider({ msal, authRequest, children }: AuthenticationContextProv
 function OidcProvider({ authRequest, children }: AuthenticationContextProviderProps) {
     // The authRequest for react-oidc is formatted slightly differently than msal.
     const oidcAuthRequest: Pick<ExtraSigninRequestArgs, 'scope' | 'extraQueryParams' | 'state'> = {
-        scope: authRequest.scopes?.join(' ') || 'openid profile email',
+        scope: authRequest.scopes?.join(' ') ?? 'openid profile email',
         extraQueryParams: authRequest.extraQueryParameters,
         state: authRequest.state,
     };


packages/context/src/authentication/AuthenticationContextProvider.tsx
```diff
@@ -64,7 +112,57 @@ function AuthenticationContextProvider(props: AuthenticationContextProviderProps
...
+                      logout: () => {
+                          auth.signoutRedirect({
+                              post_logout_redirect_uri: `/logout?p=${encodeURIComponent(
+                                  window.location.pathname + window.location.search,
+                              )}`,
+                          });
+                      },
+                      getAccessToken,
+                      isAuthenticated: true,
+                      isLoading: auth.isLoading,
```
actong, line 162: Is this ever true when the user is already authenticated?  Would it just be when logging out?

packages/context/src/authentication/AuthenticationContextProvider.tsx
```diff
@@ -64,7 +112,57 @@ function AuthenticationContextProvider(props: AuthenticationContextProviderProps
...
+                      account: { id: auth.user.profile.sub, name: auth.user.profile.name },
+                      logout: () => {
+                          auth.signoutRedirect({
+                              post_logout_redirect_uri: `/logout?p=${encodeURIComponent(
+                                  window.location.pathname + window.location.search,
+                              )}`,
+                          });
+                      },
+                      getAccessToken,
+                      isAuthenticated: true,
```
actong, line 161: Is there any value in having this as it's always true?

packages/context/src/authentication/AuthenticationContextProvider.tsx
aohsie: Hmm that's a good point, let me look into it.

packages/context/src/authentication/AuthenticationContextProvider.tsx
aohsie: Ok ya I see the point. It's supposed to indicate if the user is still authenticated but I guess we just remove the context entirely if the user logs out or the session expires so this is unnecessary.

packages/context/src/authentication/AuthenticationContextProvider.tsx
```diff
@@ -4,33 +4,81 @@
...
 import { ReactNode, createContext, useCallback, useContext, useMemo } from 'react';
+import { AuthState, useAuth } from 'react-oidc-context';
 
 export type AuthenticationContext = {
     account: UserAccount;
     logout: VoidFunction;
     getAccessToken: () => Promise<string>;
 };
 
+export type OidcAuthenticationContext = AuthenticationContext & Pick<AuthState, 'isAuthenticated' | 'isLoading'>;
```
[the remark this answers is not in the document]
aohsie: This is cleared up with the below comments.

packages/context/src/authentication/AuthenticationContextProvider.tsx
```diff
@@ -64,7 +108,55 @@ function AuthenticationContextProvider(props: AuthenticationContextProviderProps
...
+
+// OIDC Implementation
+function OidcProvider({ authRequest, children }: OidcAuthenticationContextProviderProps) {
+    const auth = useAuth();
+
+    const getAccessToken = useCallback(
+        async function () {
+            try {
+                if (auth.user?.access_token) {
+                    return auth.user.access_token;
```
actong, line 125: Does this handle expired tokens?

packages/context/src/authentication/AuthenticationContextProvider.tsx
```diff
@@ -20,17 +21,60 @@ const Context = createContext<AuthenticationContext | undefined>(undefined);
...
+// Alternative OIDC props
+export type OidcAuthenticationContextProviderProps = {
+    authRequest: OidcAuthenticationRequest;
+    children?: ReactNode;
+};
 
-function AuthenticationContextProvider(props: AuthenticationContextProviderProps) {
-    const { msal, authRequest, children } = props;
+// Combined props type for auto-detection
+export type CombinedAuthenticationContextProviderProps =
```
actong, line 46: Is this not the same as just making `msal` optional in `AuthenticationContextProviderProps`?  Is it because the types of `authRequest` are different?  Though I don't even see that react-oidc-context even makes use of that, so could make that optional too.

If you prefer having the union type, I think still keep the name `AuthenticationContextProviderProps` as the props for `AuthenticationContextProvider`, and it can be a union of `MsalAuthenticationContextProviderProps` and `OidcAuthenticationContextProviderProps`.

aohsie: I'm not sold on this approach, I did find it interesting but I'm fine doing it the other way.

packages/context/src/authentication/AuthenticationContextProvider.tsx
aohsie: Hmm I thought so but let me double check.

packages/context/src/authentication/AuthenticationContextProvider.tsx
```diff
@@ -64,7 +93,58 @@ function AuthenticationContextProvider(props: AuthenticationContextProviderProps
...
+}
+
+function OidcProvider({ authRequest, children }: OidcAuthenticationContextProviderProps) {
+    const auth = useAuth();
+
+    const getAccessToken = useCallback(
+        async function () {
+            try {
+                // Check if we have a valid (non-expired) access token
+                if (auth.user?.access_token && !auth.user.expired) {
```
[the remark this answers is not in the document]
aohsie: By default it will refresh automatically in the background (see [here](https://authts.github.io/oidc-client-ts/interfaces/UserManagerSettings.html#automaticsilentrenew). This is just a sanity check in case it didn't refresh fast enough or if `automaticSilentRenew` is set to `false` for some reason. I'll put in a better comment.

actong (approved):

PR #665, opened 2025-07-02
chore: expose account last login in authentication context

ShermanE1:
Refs: CDR-2306

### DIFF
--- a/packages/context/src/authentication/AuthenticationContextProvider.tsx
+++ b/packages/context/src/authentication/AuthenticationContextProvider.tsx
@@ -73,7 +73,11 @@ function MsalProvider({ msal, authRequest, children }: AuthenticationContextProv
         () =>
             account
                 ? {
-                      account: { id: account.localAccountId, name: account.name },
+                      account: {
+                          id: account.localAccountId,
+                          name: account.name,
+                          lastLogin: account.idTokenClaims?.last_login_time,
+                      },
                       logout: () => {
                           msal.instance.logoutRedirect({
                               account,
@@ -140,7 +144,11 @@ function OidcProvider({ authRequest, children }: AuthenticationContextProviderPr
         () =>
             auth.isAuthenticated && auth.user
                 ? {
-                      account: { id: auth.user.profile.sub, name: auth.user.profile.name },
+                      account: {
+                          id: auth.user.profile.sub,
+                          name: auth.user.profile.name,
+                          lastLogin: auth.user.profile.auth_time,
+                      },
                       logout: () => {
                           auth.signoutRedirect({
                               post_logout_redirect_uri: `/logout?p=${encodeURIComponent(


ShermanE1:
FusionAuth exposes the last login time of a user as an `auth_time` claim in the id token. This PR adds that claim to the authentication context. Additionally, a matching claim was added to the MSAL authentication context, so that the front end can use the same property from the authentication context regardless of authentication provider.

packages/context/src/authentication/AuthenticationContextProvider.tsx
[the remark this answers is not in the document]
ShermanE1: I was debating on where to put the `lastLogin` information. Do you think it makes more sense for it to be top-level rather than nested in the `account` information?

packages/context/src/authentication/AuthenticationContextProvider.tsx
```diff
@@ -73,7 +74,11 @@ function MsalProvider({ msal, authRequest, children }: AuthenticationContextProv
         () =>
             account
                 ? {
-                      account: { id: account.localAccountId, name: account.name },
+                      account: {
+                          id: account.localAccountId,
+                          name: account.name,
+                          lastLogin: account.idTokenClaims?.last_login_time,
```
aohsie, line 80: Where did you get this `last_login_time` from? It's not in the docs. I'm also getting a type error since `idTokenClaims` falls back to `[key: string]: unknown`.

packages/context/src/authentication/AuthenticationContextProvider.tsx
```diff
@@ -16,6 +16,7 @@ export type AuthenticationContext = {
 export type UserAccount = {
     id: string;
     name?: string;
+    lastLogin?: number;
```
aohsie, line 19: Should probably be `lastLoginTime`, not sure `lastLogin` is clear it's a timestamp.

packages/context/src/authentication/AuthenticationContextProvider.tsx
```diff
@@ -16,6 +16,7 @@ export type AuthenticationContext = {
 export type UserAccount = {
     id: string;
     name?: string;
+    lastLoginTime: number;
```
aohsie, line 19: This should stay optional, `auth_time` is possibly undefined.

packages/context/src/authentication/AuthenticationContextProvider.tsx
```diff
@@ -73,7 +74,11 @@ function MsalProvider({ msal, authRequest, children }: AuthenticationContextProv
         () =>
             account
                 ? {
-                      account: { id: account.localAccountId, name: account.name },
+                      account: {
+                          id: account.localAccountId,
+                          name: account.name,
+                          lastLoginTime: account.idTokenClaims?.last_login_time as number | undefined,
```
aohsie, line 80: I'm still confused where you got this from, I don't see it in the docs.

ShermanE1: It is a custom claim that is being added to the id token in the custom policies.

PR #671, opened 2025-07-15
feat(context): expose MFA session status for MSALProvider

aohsie:


ShermanE1 (approved):

actong:
What do we do today that this wasn't needed before?

aohsie:
@actong I preferred not to have to import msal into the component and instead just get the context. However, in the end I realized I still needed to get msal to initiate the actual loginRedirect.

PR #681, opened 2025-07-28
fix: only need null values at top-level properties

mtruongsa:
Refs: CDR-2202


mk160107 (approved):

mtruongsa:
> Is this different than https://github.com/Evoke-Platform/evoke-sdk/pull/682/files?

Yes. The Action and Property types still have `null` at the top-level properties. This PR is removing all `null` in nested-level properties.

mtruongsa:
> The other PR already removed them everywhere. I assume this is meant to add `null` to nested fields?

it should add `null` to top-level only. I don't know why this PR did not show no changes.

PR #684, opened 2025-07-29
fix(objects): add null to top-level fields in Property and Action types

mtruongsa:
Refs: CDR-2202


bradwray:
For my understanding, where are we explicitly setting to null? When we delete some property's property, I'd assume?

PR #686, opened 2025-08-03
feat(router): include search params in the url

mtruongsa:


packages/context/src/router/navigation.ts
```diff
@@ -2,16 +2,19 @@
...
 
 export function useNavigate(): NavigateFunction {
     const navigate = useRouterNavigate();
 
     return useCallback(
-        (page: string, params?: Record<string, string>) => {
-            navigate(generatePath(page, params));
+        (page: string, params?: Record<string, string>, searchParams?: Record<string, string>) => {
+            navigate({
+                pathname: generatePath(page, params),
```
mtruongsa, line 15: `generatePath` is used to parse params to the page url. i.e `/objects/:objectId` -> `/objects/id1`. If we have query param in the page url i.e `/objects/id1?param=1` it will become `/objects/id1param=1`

mtruongsa: nvm Configuration repo uses older react-router-dom version which has a bug.

PR #690, opened 2025-08-07
fix(CDR-2403): remove react-oidc-context from devDependencies to avoi…

aohsie:
…d version conflicts

react-oidc-context should only be a peerDependency to ensure single authentication context consuming applications can now use any compatible version (>=2) without conflicts

### DIFF
--- a/packages/context/package.json
+++ b/packages/context/package.json
@@ -56,7 +56,7 @@
         "msw": "^1.3.1",
         "react": "^18.2.0",
         "react-dom": "^18.3.1",
-        "react-oidc-context": "^2.4.0",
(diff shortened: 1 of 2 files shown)


actong:
Change itself seems fine, though I'm confused why this would cause a conflict at runtime.  Dev dependencies wouldn't be included in the final app.

actong (approved):

aohsie:
> Change itself seems fine, though I'm confused why this would cause a conflict at runtime. Dev dependencies wouldn't be included in the final app.

@actong it's not 100% clear to me but it seems it's possible since yarn hoists the dev version it might resolve to the wrong version, though that doesn't make sense to me either.

PR #695, opened 2025-08-14
fix(CDR-2403): add option to pass oidc instance to AuthenticationContext

aohsie:
There seems to be a lack of context syncing in certain edge cases. For now allow sharing the oidc instance, similar to msal, to enforce context sharing.

### DIFF
--- a/packages/context/src/authentication/AuthenticationContextProvider.tsx
+++ b/packages/context/src/authentication/AuthenticationContextProvider.tsx
@@ -5,7 +5,7 @@ import { AccountInfo, RedirectRequest } from '@azure/msal-browser';
 import { IMsalContext } from '@azure/msal-react';
 import { ExtraSigninRequestArgs } from 'oidc-client-ts';
 import { ReactNode, createContext, useCallback, useContext, useMemo } from 'react';
-import { useAuth } from 'react-oidc-context';
+import { AuthContextProps } from 'react-oidc-context';
 
 export type AuthenticationContext = {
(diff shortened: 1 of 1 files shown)


packages/context/src/authentication/AuthenticationContextProvider.tsx
```diff
@@ -100,15 +101,15 @@ function MsalProvider({ msal, authRequest, children }: AuthenticationContextProv
...
+function OidcProvider({ oidc, authRequest, children }: AuthenticationContextProviderProps) {
     // The authRequest for react-oidc is formatted slightly differently than msal.
     const oidcAuthRequest: Pick<ExtraSigninRequestArgs, 'scope' | 'extraQueryParams' | 'state'> = {
         scope: authRequest.scopes?.join(' ') ?? 'openid profile email',
         extraQueryParams: authRequest.extraQueryParameters,
         state: authRequest.state,
     };
 
-    const auth = useAuth();
+    const auth = oidc ?? useAuth();
```
actong, line 112: Maybe remove `useAuth` altogether since it has issues?

aohsie: Does it not need to be backwards compatible?

actong: `AuthenticationContextProvider` is for internal use, no widget would be using it.

actong (approved):

PR #717, opened 2025-09-25
feat: add defaultViewId to Obj type

Meisosk:
Refs: CDR-2166

### DIFF
--- a/packages/context/src/objects/objects.ts
+++ b/packages/context/src/objects/objects.ts
@@ -81,7 +81,7 @@ export type Obj = {
     properties?: Property[];
     actions?: Action[];
     formId?: string;
-    defaultViewId?: string;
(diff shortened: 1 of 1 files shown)


bradwray:
Is this something that is just for the ViewDetails widget? How will this be used and why is it needed?

Meisosk:
> et? How will this be used and why

This saves a default View on an object (unlike forms which use actions, Views have no action). It works the same way as default forms on actions meaning if a user doesn’t select a specific ViewId in the widget, it will fall back to this default. This is also needed for subtypes to show the correct View.

Meisosk:
Shlomo is working on the builder side adding Views and I'm working on the widget side. I'm also making a PR to add the defaultViewId to the object model

ShermanE1 (approved):

PR #757, opened 2026-01-16
feat: add new overload method for getInstance

mtruongsa:
Be able to send AxiosRequestConfig to the request

Refs: CDR-2557


packages/context/src/objects/objects.ts
```diff
@@ -688,13 +688,23 @@ export class ObjectStore {
...
     /**
      * Retrieves a specific instance of the object by ID.
+     * Cannot use both callback and config parameters simultaneously.
      */
     getInstance<T extends ObjectInstance = ObjectInstance>(id: string): Promise<T>;
     getInstance<T extends ObjectInstance = ObjectInstance>(id: string, cb: Callback<T>): void;
+    getInstance<T extends ObjectInstance = ObjectInstance>(
+        id: string,
+        cb?: Callback<T>,
+        config?: AxiosRequestConfig<T>,
```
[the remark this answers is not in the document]
mtruongsa: This is needed because we want to send query params to the api. We can pass `collection` and `expand` params to this endpoint `GET /objects/{id}/instances/{instanceId}`.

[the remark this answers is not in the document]
actong: `ObjectStore` is meant to abstract the api details.  If we want to provide a way to retrieve related instance data, we should add an `options` parameter that lets you declaratively specify that, shouldn't make users pass an axios request config with query parameters.

packages/context/src/objects/objects.ts
```diff
@@ -691,13 +691,34 @@ export class ObjectStore {
...
+        },
+    ) {
+        const config: AxiosRequestConfig = {
+            params: {
+                ...paramOptions,
+            },
+        };
 
-    getInstance<T extends ObjectInstance>(id: string, cb?: Callback<T>) {
         if (!cb) {
```
bradwray: I think if a they pass options as the 2nd arg, it lands in cb and gets treated as the callback

bradwray: Maybe put `|| typeof cb !== 'function'` in the if too?

packages/context/src/objects/objects.ts
```diff
@@ -689,15 +701,44 @@ export class ObjectStore {
     /**
      * Retrieves a specific instance of the object by ID.
      */
-    getInstance<T extends ObjectInstance = ObjectInstance>(id: string): Promise<T>;
+    getInstance<T extends ObjectInstance = ObjectInstance>(id: string, options?: InstanceOptions): Promise<T>;
     getInstance<T extends ObjectInstance = ObjectInstance>(id: string, cb: Callback<T>): void;
+    getInstance<T extends ObjectInstance = ObjectInstance>(
```
actong: Generally you'd also add an overload matching the function signature.  Typescript inference uses the last overload definition [see here](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#inferring-within-conditional-types:~:text=When%20inferring%20from%20a%20type%20with%20multiple%20call%20signatures%20(such%20as%20the%20type%20of%20an%20overloaded%20function)%2C%20inferences%20are%20made%20from%20the%20last%20signature%20(which%2C%20presumably%2C%20is%20the%20most%20permissive%20catch%2Dall%20case).%20It%20is%20not%20possible%20to%20perform%20overload%20resolution%20based%20on%20a%20list%20of%20argument%20types.), so without it it's going to infer `getInstance(string, Callback<T>)`.

bradwray (approved):

PR #779, opened 2026-04-22
feat(CDR-3023): dedupe concurrent get requests

smark-1:


packages/context/src/api/apiServices.ts
```diff
@@ -60,7 +85,27 @@ export class ApiServices {
             config = configOrCallback;
         }
 
-        return this.promiseOrCallback(this.api.get<T, AxiosResponse<T, D>>(url, config), cb);
+        // create unique cache key for the request based on the full URL (including baseURL) and serialized params
+        const paramStr = config?.params ? paramsSerializer(config.params) : '';
+        const cacheKey = `${this.authId}|${this.api.defaults.baseURL ?? ''}|${url}|${paramStr}`;
```
bradwray, line 95: cacheKey is based on authId|baseURL|url|params, but it ignores config.paramsSerializer. In Evoke FE we call the data-service instance endpoint with expand using repeat encoding via paramsSerializer: qs.stringify(..., { arrayFormat: 'repeat' }) at evoke-front-end/lib/ui-components/src/components/custom/Form/utils.ts:1620-1627, but other callers use axios defaults with no paramsSerializer (e.g. evoke-front-end/builder/src/services/index.ts:26), which would encode arrays as expand[]=.... Do we know for sure Cedar treats those two query shapes as equivalent, and if not, could this dedupe cause both callers to share the same in-flight promise and get the same response payload when the server would have responded differently?

bradwray: I think there’s still an Axios-supported paramsSerializer case not reflected in the cache key: paramsSerializer can be a ParamsSerializerOptions object, not just a function (AxiosRequestConfig.paramsSerializer?: ParamsSerializerOptions | CustomParamsSerializer: https://github.com/axios/axios/blob/v1.x/index.d.ts). If we don’t cover that, we could still have two requests that serialize to different query strings but share the same cache entry, so one call might get the other call’s response.

bradwray (dismissed):

packages/context/src/tests/api/api.unit.ts
```diff
@@ -486,6 +486,326 @@ describe('ApiServices', () => {
         });
     });
 
+    describe('#get request caching', () => {
+        // Each test uses a unique URL path to avoid cross-test cache contamination
+        // since the inflight maps are module-level and persist across tests.
+
```
bradwray, line 492: Since these maps live outside the class, they'll persist across every instance created in this test file. Worth adding a reset function we can call in beforeEach to clear them out?

packages/context/src/api/apiServices.ts
```diff
@@ -13,11 +13,21 @@ export type Data = Record<string, unknown> | FormData | File;
 
 const sessionId = uuidv4();
 
+const GET_CACHE_TTL_MS = 200;
+
+// Map of inflight GET requests to their promises, keyed by URL (including baseURL)
+const inflightGets = new Map<string, Promise<AxiosResponse<unknown>>>();
```
bradwray, line 19: This is a better idea.

bradwray (approved):

PR #781, opened 2026-05-04
feat: add `usePage` hook

ShermanE1:
Add a `usePage` hook that returns the page by ID. The hook must be used within the `AppProvider`.

Refs: CDR-3026

### DIFF
--- a/packages/context/src/tests/app/AppProvider.unit.tsx
+++ b/packages/context/src/tests/app/AppProvider.unit.tsx
@@ -1,0 +1,109 @@
+// Copyright (c) 2025 System Automation Corporation.
+// This file is licensed under the MIT License.
+
+import { render } from '@testing-library/react';
+import chai, { expect } from 'chai';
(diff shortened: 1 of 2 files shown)


actong (approved):

packages/context/src/tests/app/AppProvider.unit.tsx
```diff
@@ -0,0 +1,109 @@
+// Copyright (c) 2025 System Automation Corporation.
```
aohsie, line 1: Is this a new file? Maybe it should be 2026? Don't know how these things work exaclty.

aohsie: Does it even need a copywrite as a test file?

ShermanE1: I'm not sure if it is necessary, but the other test file in this repository have the copyright comment at the top of the file. 

This comment should say 2026. I can update it.

PR #784, opened 2026-05-15
fix(CDR-2881): bump lodash to 4.18.1

Meisosk:


yarn.lock
tlachman: Is the `4.17.21` version of lodash still being used by other dependencies?

Meisosk: Yeah it looks like formio still uses that version.... it's always formios fault

tlachman: It looks like the dependency versions listed above use caret (^) version ranges, so they should be able to  be updated to use v4.18.1 as well.

Meisosk: yarn upgrade lodash doesn't seem to be working for this but if I add the version here 

` "resolutions": {
        "@typescript-eslint/parser/@typescript-eslint/typescript-estree/minimatch": "^9.0.7",
        "lodash": "^4.18.1"
    },`

Then it will upgrade the versions there in the lock, should I go with this change?

tlachman (approved):

PR #790, opened 2026-06-08
chore(plugin/CDR-2943): add default theme to storybook

mtruongsa:


packages/plugin/src/generators/app/templates/.storybook/preview.ts
```diff
@@ -1,6 +1,12 @@
+import { UIThemeProvider, defaultTheme } from '@evoke-platform/ui-components';
 import type { Preview } from '@storybook/react';
+import React from 'react';
+
+const withTheme: NonNullable<Preview['decorators']>[number] = (Story) =>
```
actong: Can you not use `Decorator` instead?

mtruongsa: I updated it to use `Decorator`. The last one was Copilot suggestion.

actong (approved):

PR #791, opened 2026-07-17
feat(plugin/AIT-26): add AI coding instructions to generator scaffold

bradwray:
## What this is

This is the first time we've shipped AI coding instructions with the Evoke SDK. When a developer runs the plugin generator (`npx @evoke-platform/plugin`), they now get the option to scaffold an instruction set that teaches AI coding agents — Claude Code, Codex, or any tool that reads markdown — how to build Evoke widgets from scratch.

The core problem these instructions solve: an AI agent dropped into a plugin project doesn't know what an Evoke widget is, how `WidgetProperties.json` maps to Builder settings, which SDK hooks to use, how to verify API paths against the live OpenAPI spec, or how the Storybook mock layer works. Without this context, the agent guesses — and the guesses compile fine but silently break at runtime (wrong API paths, incorrect imports, missing MSW handlers).

## How it teaches the agent

The instruction set has two layers:

**A project-level instruction file** (`CLAUDE.md`, `AGENTS.md`, or `INSTRUCTIONS.md` depending on the developer's tool choice) that covers the fundamentals: what a plugin project is, how imports work (SDK re-exports, es5 target constraints), how to verify API endpoints against live OpenAPI specs before using them, the platform UI component library, and the Storybook/MSW conventions.

**Eleven task-specific skills** that activate when the agent is doing a particular kind of work:

- **plan-widget / build-widget** — Interview the developer to produce a blueprint, then execute it: container/presentational split, `WidgetProperties.json` authoring, SDK hook wiring, and MSW story creation
- **storybook-tdd** — Red-green TDD loop using Storybook interaction tests: write a failing `play` function first, implement until the Interactions panel goes green, then run `npm run test-storybook` for a CLI pass/fail signal
- **plan-payment-gateway / add-payment-gateway** — Safe payment gateway implementation with guardrails around sensitive financial operations
- **render-evoke-forms** — FormRendererContainer usage, action wiring, and the known change-event gap
- **build-criteria-filters** — CriteriaBuilder with the Mongo-to-Loopback operator conversion table
- **send-correspondence** — Correspondence template resolution and send endpoint usage
- **review-accessibility / review-behavioral / review-performance** — Focused code review checklists for different quality dimensions

## The environment URL prompt

The generator now optionally asks for the Evoke environment base URL at scaffold time. If provided, it's interpolated into the instruction file's "Base URL" line and immediately usable by `scripts/fetch-openapi-specs.sh` to download the environment's live API specs into `.openapi/`. This gives the agent verified endpoint data from day one instead of requiring the developer to manually edit the instruction file after scaffolding.

## Other scaffold additions

- **MSW mock layer** (`src/mocks/evokeHandlers.ts`): Pre-wired handlers for common Evoke data API patterns with a `requestLog` for asserting on submissions in Storybook play functions, plus a `resetRequestLog()` helper to prevent cross-story bleed
- **Storybook interaction testing**: `.storybook/preview.tsx` with MSW initialization and `test-storybook` npm script
- **OpenAPI spec script**: `scripts/fetch-openapi-specs.sh` downloads specs from all six Evoke service endpoints into `.openapi/` for agent reference and query-ablity via jq.

## Test plan
- [x] TypeScript compiles clean (`tsc --noEmit`)
- [x] All 12 Mocha tests pass, including 2 new `environmentUrl` interpolation tests
- [x] Lint passes (pre-commit hook)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


widgetschema.json
```diff
@@ -52,6 +52,10 @@
         "allowSticky": {
             "type": "boolean"
         },
+        "allowChildren": {
```
ShermanE1: It looks like this property was added in this commit. https://github.com/Evoke-Platform/evoke-sdk/commit/cfba304b4b1c14ddb68aeded92a5b2d5eb1e10fc

Will it be in this schema twice once this is merged in?

packages/plugin/src/generators/app/index.ts
```diff
@@ -49,8 +90,29 @@ export default class AppGenerator extends Generator {
         this.env.cwd = this.answers.dirName;
 
         this.fs.copyTpl(this.templatePath('**'), this.destinationPath(), this.answers, undefined, {
-            globOptions: { dot: true },
+            globOptions: { dot: true, ignore: ['**/_agent-instructions/**'] },
```
ShermanE1, line 94: Are `scripts/fetch-openapi-specs.sh`, the `.openapi/` entry in .gitignore, and `plans/.gitkeep` meant to be used when `agentInstructions` is 'none'? All three sit outside `_agent-instructions/`, so a 'none' scaffolded widget receives them. The script's default path exits 1 with "no CLAUDE.md / AGENTS.md / INSTRUCTIONS.md found" because that choice creates no instruction file. If they only are necessary for the agent flows, moving them into `_copyAgentInstructions` would prevent them from being included in the scaffolded widget unnecessarily.

packages/plugin/src/generators/app/templates/.gitignore
ShermanE1, line 1: This is the scaffold's first `.gitignore` and it only covers `.openapi/`, leaving every other generated path included in commits (e.g. `node_modules/`, `dist/`, `target/`, `storybook-static/`, `.storybook/public/mockServiceWorker.js` etc.). 

Should additional files or folder be added here?

packages/plugin/src/generators/app/templates/_agent-instructions/INSTRUCTIONS.md
```diff
@@ -0,0 +1,510 @@
...
+    if (status !== 'in-progress') return;
+
+    let cancelled = false;
+    const poll = async () => {
+        const result = await api.get(`/data/importRuns/${runId}`);
+        if (cancelled) return;
+        if (result.status === 'complete' || result.status === 'failed') {
+            setResult(result);
+        } else {
+            setTimeout(poll, 3000);
```
ShermanE1: Should this example call `clearTimeout`, so it isn't a "Leaked timer" as referenced here? https://github.com/Evoke-Platform/evoke-sdk/blob/e2aac49fc63cc1853ed38e3ffa923613ca4c3ae6/packages/plugin/src/generators/app/templates/_agent-instructions/skills/review-performance/SKILL.md#L30

packages/plugin/src/generators/app/templates/_agent-instructions/INSTRUCTIONS.md
```diff
@@ -0,0 +1,510 @@
...
+        }
+    };
+    poll();
```
ShermanE1: Should `poll()` have a `.catch()` here? This might give the agent mixed instructions as this line of code may fit in the "Unhandled API error" category.
https://github.com/Evoke-Platform/evoke-sdk/blob/e2aac49fc63cc1853ed38e3ffa923613ca4c3ae6/packages/plugin/src/generators/app/templates/_agent-instructions/skills/review-behavioral/SKILL.md#L35)

packages/plugin/src/generators/app/templates/_agent-instructions/skills/storybook-tdd/SKILL.md
```diff
@@ -0,0 +1,334 @@
...
+    `.storybook/public` is missing (install ran with `--ignore-scripts`), run
+    `npx msw init` once.
+-   MSW intercepts HTTP only. If a container story depends on missing React providers,
+    record that honestly and keep most red/green coverage in presentational stories.
+-   The container/Playground story should have a play function that walks through the
+    widget's primary end-to-end flow using named `step()` calls. This makes the full
+    lifecycle (e.g. upload → progress → result) visible and readable in the Interactions
+    panel.
+-   **Pace the container story for watchability.** Each named `step()` in a container
+    or end-to-end play function must begin with a 2-second delay so the developer can
```
ShermanE1, line 146: Should the `wait(2000)` be skipped under `test-storybook`, or should the the script's `--testTimeout` be raised? At 2s per step the 8-step end-to-end container story this file asks for, the test is sleeping for 16s and exceeds the runner's 15000ms default. Since the delay is for human pacing rather, it doesn't have much value in an unattended run.

packages/plugin/src/generators/app/templates/package.json
```diff
@@ -18,8 +18,10 @@
         "@storybook/addon-interactions": "^7.0.2",
         "@storybook/addon-links": "^7.0.2",
         "@storybook/blocks": "^7.0.2",
+        "@storybook/jest": "^0.2.3",
         "@storybook/react": "7.6.17",
         "@storybook/react-webpack5": "^7.0.2",
+        "@storybook/test-runner": "^0.18.0",
```
ShermanE1: Should this pin `^0.16.0`? The `@storybook/test-runner@^0.18.0` version depends on `@storybook/core-common@^8.0.0`. The runner's compatibility table maps `~0.16.0` to `^7.0.0` and `0.17.0` to `8.0.0` The `0.16.x` version looks like the line matching the scaffold's `storybook@^7.0.2`.

packages/plugin/src/generators/app/templates/package.json
```diff
@@ -46,6 +51,13 @@
         "prepackage": "npm run build",
         "package": "pkgplugin",
         "storybook": "storybook dev -p 6006",
-        "build-storybook": "storybook build"
+        "build-storybook": "storybook build",
+        "test-storybook": "test-storybook --url http://127.0.0.1:6006",
+        "postinstall": "playwright install chromium && msw init"
```
ShermanE1, line 57: Should `playwright` be a direct `devDependency` in this file? `playwright` is invoked in `postinstall`, but isn't in `devDependencies`. It is currently working as it resolved through the `@storybook/test-runner` package, Declaring it as a dependency would keep the `postinstall` from depending on another package's dependency tree.

packages/plugin/src/generators/app/templates/src/mocks/evokeHandlers.ts
```diff
@@ -0,0 +1,112 @@
...
+// real SDK client. When the Storybook preview console warns about an unhandled
+// request, add a handler for that exact path here. MSW matches the path without query
+// strings, so mock `/api/data/objects/:objectId/instances` rather than trying to embed
+// `?filter=...` in the handler path.
+import { HttpResponse, http } from 'msw';
+
+// Capture submissions so play functions can assert on what the widget actually sent.
+// Call resetRequestLog() in each story's play function before interacting, so
+// submissions from prior stories don't leak into assertions.
+export const requestLog: { submissions: Array<{ path: string; body: unknown }> } = {
```
ShermanE1, line 27: Should `preview.tsx` `loaders` call `resetRequestLog()`, so this doesn't rely on every author calling it between stories? Loaders run per story before render, so `loaders: [mswLoader, () => { resetRequestLog(); }]` would ensure the `requestLog` is cleared between stories.

widgetschema.json
bradwray: good catch. I ran the merge and it does come out twice, ours at line 55 and main's at 71. it merges clean because the two additions sit in different places, so nothing flags it as a conflict. dropping our block and moving our description onto main's entry.

packages/plugin/src/generators/app/index.ts
bradwray: you're right, all three land in a 'none' scaffold today. moving `scripts/fetch-openapi-specs.sh` and `plans/` into `_copyAgentInstructions` so they only ship when an agent choice is made. keeping `.gitignore` universal since every project wants one.

packages/plugin/src/generators/app/templates/.gitignore
bradwray: yeah, this only covers `.openapi/`. adding `node_modules/`, `dist/`, `target/`, `storybook-static/`, `.storybook/public/`, `coverage/`, `*.log`, `.DS_Store`.

packages/plugin/src/generators/app/templates/_agent-instructions/INSTRUCTIONS.md
bradwray: correct, and it's our own review-performance skill flagging it. capturing the timer id and clearing it in the cleanup so the example passes the rule it teaches.

packages/plugin/src/generators/app/templates/_agent-instructions/INSTRUCTIONS.md
bradwray: agreed, that's the "Unhandled API error" row in review-behavioral. adding a `.catch()` that moves to the error phase, so the example matches the fix direction the skill asks for.

packages/plugin/src/generators/app/templates/_agent-instructions/skills/storybook-tdd/SKILL.md
bradwray: right, eight steps at 2s each overruns the runner. making the wait a no-op when `navigator.webdriver` is set, so it still paces in the browser but costs nothing under test-storybook.

packages/plugin/src/generators/app/templates/package.json
bradwray: confirmed: 0.16.0 pulls `core-common@^7`, 0.17 and up pull `^8`. requiring 0.18 locally actually fails with "Cannot find module '@storybook/core-common'". pinning `^0.16.0` to match storybook 7.

packages/plugin/src/generators/app/templates/package.json
bradwray: agreed, it's only resolving through the test-runner's tree right now. adding `playwright` as a direct devDependency, which matters more given we're changing that tree anyway.

packages/plugin/src/generators/app/templates/src/mocks/evokeHandlers.ts
bradwray: good idea, better than trusting every author to remember. wiring `resetRequestLog()` into the `preview.tsx` loaders and dropping the instruction from the comment.

ShermanE1 (approved):

packages/plugin/src/__tests__/create-plugin.ts
```diff
@@ -8,18 +8,231 @@ import helpers, { RunResult } from 'yeoman-test';
...
+            .withPrompts({ projectName: 'test', dirName: 'testdir', agentInstructions: 'claude' });
+
+        // A generated project is not a git repo, so the test-runner's getProjectRoot()
+        // walk escapes it and lands on the user's home directory — jest then scans
+        // everything under $HOME and the run appears to hang.
+        runResult.assertFile(['testdir/test-runner-jest.config.js']);
+        runResult.assertFileContent('testdir/test-runner-jest.config.js', 'rootDir: __dirname');
+    }).timeout(5000);
+
+    it('keeps the storybook preview free of the two defects that break every story', async () => {
```
ShermanE1, line 174: These three assertions pin the exact strings — is it worth closing the class too? A different bad named import in `preview.tsx` still ships, and `storySort: (a, b: {...})` slips past the `storySort: (a: ` literal.

Adding `.storybook` to tsconfig's `include` with a `"typecheck": "tsc --noEmit"` in the validation gate catches it: under the scaffold's own compilerOptions `tsc` reports `TS2305: Module '@evoke-platform/ui-components' has no exported member 'UIThemeProvider'`.

bradwray: You're right that those three only catch the exact spellings. Worth checking before you rely on it though: listing `.storybook` in `include` matched zero files for me, since TypeScript skips dot-prefixed folders unless you write `.storybook/**/*`.

With that corrected it does report TS2305 on both names, but it also flags implicit-any on the `storySort` params, which are untyped deliberately: Storybook re-parses that function as plain JavaScript and throws on any annotation. Would you add the typecheck alongside these assertions with a suppression there, or swap the literal match for a test that runs getStorySortParameter on the generated file?

ShermanE1: I'm fine with either. The advantage of the second approach is that it tests what Storybook actually does with the file rather than one spelling of the mistake: annotating only `b`, or the inner `rank` helper, or an extra space before the colon all slip past the literal, and all of them throw in the parser. It also fails with the same error the index builder raises at runtime, so the test names the real cause. The cost is a new devDependency on `@storybook/csf-tools` for one test, which may not be worth it.

bradwray: Went with the parser approach, and it turned out not to need the dependency. The test pulls the storySort source out of the generated preview and does `new Function('return ' + src)`, which fails on exactly the same syntax for the same reason. I checked it against `@storybook/csf-tools` on five cases including the three you named and it gave identical verdicts, then reintroduced each variant into the template to confirm the test actually goes red: annotating only `b`, the inner `rank` helper, and a space before the colon are all caught.

Pushed in 71d7de0. Left the two import assertions as they are, since those are the literal strings that shipped.

ShermanE1 (dismissed):

packages/plugin/README.md
```diff
@@ -25,6 +25,22 @@ npm run package
 A deployable zip will be created in the `target/` directory under the project root, which can be uploaded to
 an Evoke environment.
 
+### Agent Guidance
+
+The generator can optionally scaffold AI coding instructions.
+
+When prompted, choose:
+
+-   **Claude Code (recommended):** adds `CLAUDE.md` and eleven skills under `.claude/skills/`.
```
actong, line 34: Why is this the recommended one?

packages/plugin/README.md
```diff
@@ -25,6 +25,22 @@ npm run package
...
+-   **Codex:** adds `AGENTS.md` and eleven skills under `.agents/skills/`.
+-   **Generic instructions only:** adds `INSTRUCTIONS.md` and eleven skills under `.agents/skills/`
+    (readable as plain documentation by tools without a skill mechanism).
+-   **No AI instructions:** adds no agent files.
+
+The generated guidance covers project structure, commands, widget configuration, forms,
+test-first Storybook development, criteria filters, correspondence sending, payment gateway safety, and guardrails.
+Generated projects include Storybook interaction testing with `npm run test-storybook` for a CLI pass/fail signal.
```
actong: This seems out of place. What does it have to do with agent guidance?

packages/plugin/src/generators/app/index.ts
```diff
@@ -5,9 +5,25 @@ import chalk from 'chalk';
...
     projectName: string;
     dirName: string;
+    agentInstructions: AgentInstructions;
+    environmentUrl: string;
+};
+
+const instructionFileNames: Record<Exclude<AgentInstructions, 'none'>, string> = {
+    claude: 'CLAUDE.md',
+    codex: 'AGENTS.md',
+    generic: 'INSTRUCTIONS.md',
```
actong: What clients use this?

packages/plugin/src/generators/app/templates/_agent-instructions/skills/add-payment-gateway/SKILL.md
```diff
@@ -0,0 +1,100 @@
...
+ * @paymentGateway
+ */
+export class MyGateway implements PaymentGateway {
+    constructor(private props: MyGatewayProps) {}
+
+    prepare(payment: Payment, returnUrl: string): TransferData | PromiseLike<TransferData> {
+        throw new Error('Method not implemented.');
+    }
+
+    postPaymentResult(payment: Payment, resultData: ParsedQueryString): Payment | null | PromiseLike<Payment | null> {
```
actong, line 45: Should probably mention this and `receivePaymentNotification` need to perform their own auth validation, which would vary from provider to provider. Otherwise anyone can fake a payment.

packages/plugin/src/generators/app/templates/_agent-instructions/skills/build-criteria-filters/SKILL.md
```diff
@@ -0,0 +1,69 @@
...
+---
+
+# Build Criteria Filters in a Widget
+
+Use `CriteriaBuilder` when the widget needs a Builder-style filter editor. It edits Evoke's
+Mongo-style criteria objects, the same shape used in `filter` widget properties.
+
+Before implementing, inspect the installed package for the current props and examples:
+
+-   `node_modules/@evoke-platform/ui-components/dist/published/components/custom/CriteriaBuilder/CriteriaBuilder.d.ts`
```
actong: This hardcodes a dependence on specific paths that could change in the future

packages/plugin/README.md
bradwray: The instructions were mostly written and tested against Claude Code, and Codex got fewer runs. I have run it with Codex and it does work. Whether that justifies the word "recommended" is a judgement call. I can take it out.

packages/plugin/README.md
bradwray: It's there because the agent instructions depend on it. An agent writing UI code can't see what it built, so `npm run test-storybook` returning a plain pass or fail is what closes the loop. In a dogfood run last week that's what caught a DataGrid overflowing its container by 16px, which reading the code would never have surfaced.

That said, the sentence as written says none of that, which is why it reads as out of place. Would this work better?

> The agent instructions are built around a feedback loop. Generated projects include Storybook interaction tests and `npm run test-storybook`, which returns a plain pass or fail. Without that, an agent is changing UI code blind, with no way to tell what actually worked.

The testing itself ships in every generated project regardless of which agent option you pick, so if you'd rather this section stay strictly about agent guidance I can move it up to the scaffold description instead.

packages/plugin/src/generators/app/index.ts
bradwray: None. Claude Code reads `CLAUDE.md` and `AGENTS.md` is the cross-tool convention, so this one was just the Codex option with a filename nothing loads.

Removed it in b754822.

packages/plugin/src/generators/app/templates/_agent-instructions/skills/add-payment-gateway/SKILL.md
bradwray: Agreed, and `postPaymentResult` wasn't covered anywhere: everything we had was about webhooks. I'm marking both methods in the skeleton as carrying data anyone can forge, saying it again where the skill explains how to record the result, and adding a line to the main instruction file.

packages/plugin/src/generators/app/templates/_agent-instructions/skills/build-criteria-filters/SKILL.md
bradwray: Good catch, and it was 21 paths across four files rather than just this one. Swapped them all for searches, which also turned out to return four files the criteria list was missing.

In a369ee1.

packages/plugin/src/generators/app/templates/_agent-instructions/skills/add-payment-gateway/SKILL.md
```diff
@@ -85,15 +120,24 @@ These fields are defined on the `Payment` interface in `@evoke-platform/sdk` —
 type as the source of truth. Inspect the installed declarations for the exact current
 shapes:
 
--   `node_modules/@evoke-platform/payment/dist/payment.d.ts`
--   `node_modules/@evoke-platform/payment/dist/paymentGateway.d.ts`
+```bash
+find node_modules/@evoke-platform/payment -name 'payment.d.ts' -o -name 'paymentGateway.d.ts'
```
actong: This still relies on specific file names. Why can't it just get the type declarations from the package `@evoke-platform/payment` (the typings should be exported from the package) instead of telling it to look for specific file names?

packages/plugin/README.md
actong: Yeah, at least it wouldn't look out of place.

packages/plugin/src/generators/app/templates/_agent-instructions/skills/add-payment-gateway/SKILL.md
bradwray: Simplified this to point directly at the installed packages’ exported types. Removed the filename searches and lookup procedure.

packages/plugin/src/generators/app/templates/_agent-instructions/skills/add-payment-gateway/SKILL.md
bradwray: Updated the planning checklist and skeleton so both callbacks explicitly require authentication using the provider’s documented mechanism.

packages/plugin/README.md
bradwray: Done in f9667a5.

actong:
I don't know if this changes what you want to do, but Anthropic added support for AGENTS.md.

PR #794, opened 2026-08-05
feat(CDR-3323): support valuePath for nested API options in widget api schema

Meisosk:


widgetschema.json
```diff
@@ -110,13 +110,17 @@
         },
         "api": {
             "type": "object",
-            "required": ["url", "resultMapping"],
+            "required": ["url"],
```
Meisosk, line 113: resultMapping was required on api, but multi-select fields don't use it. Made it optional so valid multi-select configs using just api.valuePath pass schema validation instead of failing

mtruongsa (approved):

